PR #1, opened 2015-11-03
Qingwei/desktop deploy

qingweibinary:
I create two deploy task, one for bash terminal, one for windows cmd, the latter is not tested

The first one works, but we can improve it's output


npm_tasks/gh-pages-win.bat
```diff
@@ -0,0 +1,21 @@
...
+
+git checkout gh-pages
+git checkout %CURRENT_BRANCH -- dist/*
+git checkout %CURRENT_BRANCH -- public/*
+
+git add dist/*
+git add public/*
+
+git commit -m "deployed from %CURRENT_BRANCH"
+git push --force origin gh-pages
```
qingweibinary: Here is using force push, which might not be appropriate, I am still looking for answer, kindly comment ~

npm_tasks/gh-pages.sh
```diff
@@ -0,0 +1,21 @@
...
+
+git checkout gh-pages
+git checkout $CURRENT_BRANCH -- dist/*
+git checkout $CURRENT_BRANCH -- public/*
+
+git add dist/*
+git add public/*
+
+git commit -m "deployed from $CURRENT_BRANCH"
+git push --force origin gh-pages
```
qingweibinary: Here is using force, which might not be appropriate, I am still looking for answer, kindly comment ~

PR #4, opened 2015-11-05
Qingwei/hello world routing

qingweibinary:
Still not able to render the view ~
not sure if I understand the routing correctly

### DIFF
--- a/src/_reducers/HelloWorldReducer.js
+++ b/src/_reducers/HelloWorldReducer.js
@@ -1,0 +1,15 @@
+import StateStorage from '../_store/StateStorage';
+
+const SayHello = 'Hello';
+const SayHi = 'Hi';
+const NoSayHello = '';
+
+const initialState = new Map({subject: 'World'});
+
+export default (state=initialState, action=null) => {
+    switch (action.type) {
+        case SayHello: return state.set('greet', 'Hello');
+        case NoSayHello: return state.set('greet', '');
+        case SayHi: return state.set('greet', 'Hi');
+    }
+}
--- a/src/_routes/index.js
+++ b/src/_routes/index.js
@@ -4,7 +4,9 @@ import TickTradeRoutes from './TickTradeRoutes';
 import DesktopRoutes from './DesktopRoutes';
 import CommonRoutes from './CommonRoutes';
 
-const childRoutes = [].concat(DesktopRoutes).concat(TickTradeRoutes).concat(CommonRoutes);
+import HelloWorldRoute from './HelloWorldRoute'
+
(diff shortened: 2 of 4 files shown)


src/_reducers/HelloWorldReducer.js
```diff
@@ -0,0 +1,15 @@
+import StateStorage from '../_store/StateStorage';
```
borisyankov: This is a custom built thing by me, we should be replacing it with redux-localstorage (which seems to be better).

src/_reducers/HelloWorldReducer.js
```diff
@@ -0,0 +1,15 @@
...
+
+const SayHello = 'Hello';
+const SayHi = 'Hi';
+const NoSayHello = '';
+
+const initialState = new Map({subject: 'World'});
```
borisyankov: You need to import 'immutable' or else you are using the ES6 Map. Which will lead to unexpected behavior ;) One solution is to always use Immutable.Map, another is to continue being cool and using it like this but being mindful. ;)

src/_reducers/HelloWorldReducer.js
```diff
@@ -0,0 +1,15 @@
...
+
+export default (state=initialState, action=null) => {
```
borisyankov: I don't think action = null is necessary, the default will be undefinied which is almost equivalent.
The rest looks correct.

src/_reducers/HelloWorldReducer.js
qingweibinary: ok, now I get what you meant previously

src/_routes/index.js
```diff
@@ -4,7 +4,9 @@ import TickTradeRoutes from './TickTradeRoutes';
 import DesktopRoutes from './DesktopRoutes';
 import CommonRoutes from './CommonRoutes';
 
-const childRoutes = [].concat(DesktopRoutes).concat(TickTradeRoutes).concat(CommonRoutes);
+import HelloWorldRoute from './HelloWorldRoute'
+
+const childRoutes = [].concat(DesktopRoutes).concat(TickTradeRoutes).concat(CommonRoutes).concat(HelloWorldRoute);
```
borisyankov: You can't concat an object to an array, you would have to do .concat([HelloWorldRoute])

src/helloworld/HelloWorldPanel.js
```diff
@@ -0,0 +1,19 @@
...
+        subject: String
+    }
+
+    render() {
+        return
+        <div>
+            <p>{this.props.greet + " " + this.props.subject}</p>
+        </div>
+    }
+}
```
borisyankov: Looks correct

src/_reducers/HelloWorldReducer.js
qingweibinary: yea, thanks for catching

PR #5, opened 2015-11-09
Youtube video

qingweibinary:
There's a hardcoded video now, next I will try to fetch video url by user account, if it's not feasible I will then hardcode the whole list of videos into our code

The route is at

http://localhost:3000/#/video


src/video/VideoPlayer.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import { connect } from 'react-redux';
+import YouTube from 'react-youtube';
+
+@connect(state => ({urls: state.urls, idx: state.currentIndex}))
```
borisyankov: That part here is likely not working.
You need to write it as {videos: state.videos}

qingweibinary: argh, so this is the problem, thanks

src/_constants/ActionTypes.js
```diff
@@ -40,3 +40,6 @@ export const SIGNIN_FAILED = 'SIGNIN_FAILED';
 
 export const SERVER_DATA_ACCOUNT_LIMITS = 'SERVER_DATA_ACCOUNT_LIMITS';
 export const SERVER_DATA_ACCOUNT_SETTINGS = 'SERVER_DATA_ACCOUNT_SETTINGS';
+
+export const NEXT_VIDEO = 'NEXT_VIDEO';
+export const PREVIOUS_VIDEO = 'PREVIOUS_VIDEO';
```
borisyankov: This would be just a single action 'CHANGE_SELECTED_VIDEO'.
And then it can be reused even for arbitrary clicking on a video in a list of videos.

src/_reducers/VideoReducers.js
```diff
@@ -0,0 +1,27 @@
+import { Map } from 'immutable';
+import { NEXT_VIDEO, PREVIOUS_VIDEO } from '../_constants/ActionTypes'
+
+const initialState = new Map({
+    urls: ['https://www.youtube.com/watch?v=zQJy9sFHU38'],
```
borisyankov: May be you need just the 'IDs' like 'zQJy9sFHU38' ;)

borisyankov:
Good. Keep working on it ;)

src/_reducers/VideoReducers.js
qingweibinary: will test

src/_constants/ActionTypes.js
qingweibinary: hmm, so now we do not want to have the video in the apps? and user can swipe left or right to view next?

borisyankov: Oh, we will ;) 
Just the way we notify the change internally will be different.

You will also need a <VideoList> component, custom created by you, just reuse the HTML I generate in the blog itself. Will be enough for now.

src/video/VideoList.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import { connect } from 'react-redux';
+import VideoRow from './VideoRow.js'
+
+const VideoList = (videos) => (
+    <ul>
+        { videos.map((v) =>
```
qingweibinary: @borisyankov here gives me error where videos does not have map function, did I pass it wrongly?

qingweibinary: I figured it out ....

borisyankov: You get passed 'props' and you can destructure it.
So it should be ({videos})

src/_constants/ActionTypes.js
```diff
@@ -40,3 +40,7 @@ export const SIGNIN_FAILED = 'SIGNIN_FAILED';
 
 export const SERVER_DATA_ACCOUNT_LIMITS = 'SERVER_DATA_ACCOUNT_LIMITS';
 export const SERVER_DATA_ACCOUNT_SETTINGS = 'SERVER_DATA_ACCOUNT_SETTINGS';
+
+export const CHANGE_VIDEO = 'CHANGE_VIDEO';
```
borisyankov: I'd call it CHANGE_ACTIVE_VIDEO

qingweibinary: will fix

src/video/VideoRow.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+import { connect } from 'react-redux';
+import * as VideoActions from '../_actions/VideoActions';
+
+
+@connect()
```
borisyankov: You will not need that ;)

src/video/VideoRow.js
```diff
@@ -0,0 +1,23 @@
...
+export default class VideoRow extends React.Component {
+
+    onClick(){
+        this.props.dispatch(VideoActions.changeVideo(this.props.url));
```
borisyankov: This will be handled in the parent, so this component is simpler.
Pass onSelect prop to this, and onClick={onSelect}
Then dispatch (or actually the raw action prop will be called from parent)

src/video/VideoPanel.js
```diff
@@ -0,0 +1,44 @@
...
+import VideoData from '../_data/VideoData';
+import * as VideoAction from '../_actions/VideoActions';
+
+@connect(state => ({video: state.video}))
+export default class VideoPanel extends React.Component {
+
+    componentDidMount() {
+        VideoData.getAll().then((response) => {
+            return response.json();
+        }).then((jsonObj) => {
```
borisyankov: Do you need to chain two Promises? 
You can handle it in the first one?

src/video/VideoPanel.js
```diff
@@ -0,0 +1,44 @@
...
+        const videos = this.props.video.get('videos');
+
+        return (
+            <div>
+                <Panel title={title}>
+                    <YouTube url={url}/>
+                </Panel>
+                <Panel title={"Video List"}>
+                    <VideoList videos={videos} />
+                </Panel>
```
borisyankov: Not sure we will do it with two Panels per component, but since I'll be reworking it anyways, we'll leave it like that.

src/video/VideoPanel.js
qingweibinary: I will try, previously I seems to have some problem

src/video/VideoRow.js
qingweibinary: I'll try fix this

src/video/VideoPanel.js
qingweibinary: response.json() returns a promise, thus we need to chain it

PR #12, opened 2015-11-26
Qingwei/mobile improvement

qingweibinary:


src/_actions/SettingsActions.js
```diff
@@ -19,3 +19,8 @@ export const serverDataAccountSettings = serverResponse => ({
     type: types.SERVER_DATA_ACCOUNT_SETTINGS,
     serverResponse,
 });
+
+export const updatePersonalAddress = newAddress => ({
```
borisyankov: I had a more generic action elsewhere.
I suggest it for settings too:
LIke UPDATE_SETTINGS_FIELD, it will have a fieldName/fieldKey and fieldValue params, so you use the same action for all settings.

src/_common/LabeledText.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+
+const LabeledText = ({id, label, value = 'N/A'}) => (
+    <tr id={id} className={'name-val-pair'}>
```
borisyankov: Flexbox is more powerful than tables.
I would have created this as a pair of label elements, or span elements, then style them.

src/_routes/DesktopRoutes.js
```diff
@@ -31,27 +31,27 @@ import { requireAuthOnEnter } from '../_data/Auth';
...
-    { path: 'daily-prices', component: DailyPricesPage },
-    { path: 'intraday-prices', component: IntradayPricesPane },
-    { path: 'profit-table', component: ProfitTablePage },
-    { path: 'settings', indexRoute: { component: SettingsPersonalDetails }, component: SettingsPage, childRoutes: [
+    { path: 'rise-fall-table', component: RiseFallTablePage, onEnter: requireAuthOnEnter },
+    { path: 'pricing-table', component: PricingTablePage, onEnter: requireAuthOnEnter },
+    { path: 'daily-prices', component: DailyPricesPage, onEnter: requireAuthOnEnter },
+    { path: 'intraday-prices', component: IntradayPricesPane, onEnter: requireAuthOnEnter },
+    { path: 'profit-table', component: ProfitTablePage, onEnter: requireAuthOnEnter },
+    { path: 'settings', indexRoute: { component: SettingsPersonalDetails }, onEnter: requireAuthOnEnter, component: SettingsPage, childRoutes: [
```
borisyankov: Good stuff, I noticed we are missing these ;)

src/settings/SettingsDetails.js
```diff
@@ -9,36 +10,35 @@ export default class SettingsDetails extends React.Component {
...
-					value={settings.country}
-					readOnly />
-				<InputGroup
-					id="email"
-					label="Email"
-					type="email"
-					value={settings.email}
-					readOnly />
-
+				<LabeledTextList
```
borisyankov: This component was slightly over engineered ;)
Just adding several LabeledText components one after the other would have a bit more html-ish feel and be more configurable than having a LabeledTextList ;)

src/_actions/SettingsActions.js
qingweibinary: Ah, did not notice about it~
Will fix

src/_common/LabeledText.js
qingweibinary: ok, I'll change it to something more generic that works with flex

src/settings/SettingsDetails.js
qingweibinary: I did it this way because I need something as a parent component which can control the alignment of children component, children cannot be independent or the alignment will be broken

is there any solution on this?

borisyankov: Pure CSS can control that reasonably well ;)
Also will be more responsive, in theory.
If you set a flex: 1 to say .label-label and flex: 2 to say .label-value it will always be 1 to 2 per line (especially if wrapped in fieldset or some other tag)

qingweibinary: OK, I'll change this then

src/_actions/SettingsActions.js
qingweibinary: @borisyankov I only able to find `serverDataAccountSettings`, which I ignore it at first because semantically it perform something else (update store from server data) rather than (update store) which is more generic, so do you think I should create a generic action like what u mentioned?

PR #14, opened 2015-11-27
Qingwei/improve perf

qingweibinary:
Improve the performance of Tabs
Problem: 
- active tab is stored in store, to update it, an action is fired, which is not necessary, solve by using Component state.


borisyankov:
Not sure we should do it this way, but will merge and fix from there.

PR #17, opened 2015-11-30
Tick sparkline now start when initial load

qingweibinary:


src/_data/LiveData.js
```diff
@@ -28,6 +28,13 @@ const handlers = {
 
 export const api = new LiveApi({ language: 'EN' });
 
+const subscribeToSelectedSymbol = () => {
+    loadedStorePromise.then(st => {
```
qingweibinary: Not sure if we want to hardcode or not? as the initial symbol is hardcoded

borisyankov: LiveData should not be aware of the store that much, this needs to be injected from parent.

qingweibinary: Yes, I note this in the `Refactor` todo, do we want to do it now?
I forgot why I did not refactor it directly, I believe because a few places need to be change

borisyankov: Then add in the mental to-do-next-after-release-to-app-stores.

PR #18, opened 2015-11-30
Qingwei/show tick trade only

qingweibinary:
This depends on https://github.com/binary-com/binary-next-gen/commit/b358e37d8bf3703e9fdaa533756e4afc17fab212

AssetSelectors and Assets state need to access each other in order to obtain sufficient information to filter out Non-tick trade, result in some useless state and ugly logic used as a workaround.

To solve it elegantly we might need some redesign on redux store shape, now we mainly store whatever we get from api.

### DIFF
--- a/src/_data/LiveData.js
+++ b/src/_data/LiveData.js
@@ -47,9 +47,9 @@ const subscribeToWatchlist = () => {
 };
 
 export const initUnauthorized = () => {
-    api.getActiveSymbolsFull();
     api.getTradingTimes();
     api.getAssetIndex();
+    api.getActiveSymbolsFull();
 
     readNewsFeed().then(articles => api.events.emit('news', articles));
     getVideosFromPlayList().then(videos => api.events.emit('videos', videos));
--- a/src/_reducers/AssetSelectorReducers.js
+++ b/src/_reducers/AssetSelectorReducers.js
@@ -28,9 +28,9 @@ const doFilter = (AssetSelectorList, query, markets, submarket) => {
     ).sort((x1, x2) => x1.get('display_name').localeCompare(x2.get('display_name')));
 };
 
-const hasTick = underlyings => {
-    const withTicks = underlyings.filter(underlying => {
-        return underlying[2].includes('t');
+const hasTick = assets => {
+    const withTicks = assets.filter(asset => {
+        return asset[2].includes('t');
     });
     return withTicks.length > 0;
 };
@@ -38,9 +38,7 @@ const hasTick = underlyings => {
 const tickTradeFilter = assetIndex => {
     const symbolWithTick = assetIndex.filter(asset => {
         return hasTick(asset[2]);
-    }).map(asset => {
-        return asset[0];
-    });
+    }).map(asset => asset[0]);
     return symbolWithTick;
 };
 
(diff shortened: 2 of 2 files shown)


src/_data/LiveData.js
```diff
@@ -40,17 +47,18 @@ const subscribeToWatchlist = () => {
 };
 
 export const initUnauthorized = () => {
-    api.getActiveSymbolsFull();
     api.getTradingTimes();
     api.getAssetIndex();
 
     readNewsFeed().then(articles => api.events.emit('news', articles));
     getVideosFromPlayList().then(videos => api.events.emit('videos', videos));
+    subscribeToSelectedSymbol();
```
borisyankov: This should not happen here.
LiveData should have no idea what is selected symbol or that we even use immutable.js :dancer:

src/_reducers/AssetSelectorReducers.js
```diff
@@ -27,11 +28,30 @@ const doFilter = (AssetSelectorList, query, markets, submarket) => {
     ).sort((x1, x2) => x1.get('display_name').localeCompare(x2.get('display_name')));
 };
 
+const hasTick = underlyings => {
+    const withTicks = underlyings.filter(underlying => {
```
borisyankov: Lets just use 'assets' for consistency everywhere.

src/_reducers/AssetSelectorReducers.js
```diff
@@ -27,11 +28,30 @@ const doFilter = (AssetSelectorList, query, markets, submarket) => {
...
+        return underlying[2].includes('t');
+    });
+    return withTicks.length > 0;
+};
+
+const tickTradeFilter = assetIndex => {
+    const symbolWithTick = assetIndex.filter(asset => {
+        return hasTick(asset[2]);
+    }).map(asset => {
+        return asset[0];
```
borisyankov: This can be written as .map(asset => asset[0]);

src/_actions/WorkspaceActions.js
```diff
@@ -55,3 +55,11 @@ export const delFromWatchlist = symbol => {
         dispatch(workspaceUnfavorAsset(symbol));
     };
 };
+
+export const selectAssetSymbolForTrade = (newSymbol, oldSymbol) => {
+    return dispatch => {
+        LiveData.api.unsubscribeFromTick(oldSymbol);
+        LiveData.api.subscribeToTick(newSymbol);
```
borisyankov: I was just thinking about your reducers as pure functions, and you are digging us deeper ;)
Just keep in mind for next refactoring!

qingweibinary: Sorry, what do you mean by digging deeper?

borisyankov: As we should not have calls to the api from an action. We'll see how to solve that.

qingweibinary: well, I follow recommendation from the creator

https://github.com/rackt/redux/issues/291

any reason not to follow? or do you mean we should abstract it further?

borisyankov: Yeah, we'll need to abstract it a bit, but haven't read this thread. Good find ;)

PR #21, opened 2015-12-01
Fix asset selector problem

qingweibinary:


src/_store/Root.dev.js
```diff
@@ -9,8 +9,8 @@ import * as LiveData from '../_data/LiveData';
 
 const history = new HashHistory();
 rehydratedStorePromise.then(st => {
-    const token = st.getState().signin.get('token');
-    LiveData.connect(st, token);
+    // const token = st.getState().signin.get('token');
+    LiveData.connect(st, '');
```
qingweibinary: here shows that we do not have to use connect to authorize, but then the flow would be less obvious

`connect` attach call back, and requireOnAuth does the auth api call

PR #30, opened 2015-12-03
Qingwei/translation

qingweibinary:


src/_store/Root.dev.js
```diff
@@ -14,6 +14,7 @@ rehydratedStorePromise.then(st => {
 
 export default class Root extends Component {
     render() {
+        // require a huge pile of text file deserialize into JS obj for translation
```
borisyankov: Doesn't need to be in the render() method, can be on line 14 for example

qingweibinary: yea, I was actually hoping to have a cdn of js-object ~

src/portfolio/PortfolioTable.js
```diff
@@ -12,9 +13,27 @@ const PortfolioTable = ({compact, contracts, proposals, onViewDetails}) => {
 		<table>
 			<thead>
 				<tr>
-					<th>Ref.</th>
-					<th>Purchase</th>
-                    <th>Indicative</th>
+					<th>
+						<FormattedMessage
+							id="ref"
+							description=""
```
borisyankov: Is id mandatory? And not sure if we want to add description here... if it is empty we don't, if it isn't... need to think it through ;)

src/signin/SigninCard.js
```diff
@@ -72,10 +72,28 @@ export default class SigninCard extends React.Component {
...
+					<FormattedMessage
+						id="sign-in"
+						description="button for sign in"
+						defaultMessage="Sign In"
+						/>
+				</button>
+				<a className="outline-link" target="new" href="https://www.binary.com/user/api_token">
+					<FormattedMessage
+						id="get-token"
+						description="button to get api token"
```
borisyankov: Not sure these descs add anything useful, may be use as exception than the rule

src/signin/SigninCard.js
```diff
@@ -72,10 +72,28 @@ export default class SigninCard extends React.Component {
 					text="Access denied" />
 
 				<LanguagePicker onChange={::this.onLanguageChange} />
-				<button className="outline-link signin-btn" onClick={::this.trySignin}>Sign In</button>
-				<a className="outline-link" target="new" href="https://www.binary.com/user/api_token">Get your API token</a>
+				<button className="outline-link signin-btn" onClick={::this.trySignin}>
+					<FormattedMessage
+						id="sign-in"
+						description="button for sign in"
```
borisyankov: Possibly reused for other stuff, not just buttons. I think 'Sign in' is clear enough. May be a description in theory would be 'used to sign in the user into an account' so translators know the context. I'd say leave all empty for now. No 'description' property at all.

src/portfolio/PortfolioTable.js
qingweibinary: i think id is needed, description is probably not, plan to clear all of them after I confirm weblate does not use it

src/watchlist/TickTradeSparkline.js
```diff
@@ -8,18 +8,20 @@ const ypos = (data, val) => ((val - min(data)) / (max(data) - min(data)) * 120);
...
+	const widthPerTick = 18;
+
 	return (
 		<div className="tick-sparklines">
 			<Sparklines
 				width={width}
 				height={height}
 				data={h}
-				limit={20}
+				limit={Math.round(width / widthPerTick)}
```
borisyankov: Possibly still a good idea to limit to 20.

src/watchlist/TickTradeSparkline.js
```diff
@@ -8,18 +8,20 @@ const ypos = (data, val) => ((val - min(data)) / (max(data) - min(data)) * 120);
 const history = data => data.map(x => x.quote);
 
 const TickTradeSparkline = (props) => {
-	const height = 100;
-	const width = 344;
+	const height = props.height || 100;
```
borisyankov: Add TickTradeSparkline.defaultProps = {} instead of this.

src/watchlist/TickTradeSparkline.js
```diff
@@ -37,6 +39,8 @@ const TickTradeSparkline = (props) => {
 TickTradeSparkline.propTypes = {
 	history: React.PropTypes.array,
 	showBarrier: React.PropTypes.bool,
+	width: React.PropTypes.number,
+	height: React.PropTypes.number,
 };
 
 TickTradeSparkline.defaultProps = {
```
borisyankov: Here it is ;)

src/watchlist/TickTradeSparkline.js
qingweibinary: opps, this is an accidental commit, but as screen gets bigger, dont we want to show more data?

borisyankov: Not sure, not necessarily.
So the answer can be both yes and no ;)

qingweibinary:
this PR is not completed as the translated strings are not injected yet

src/profit-table/ProfitTable.js
```diff
@@ -1,19 +1,31 @@
 import React from 'react';
 import { NumberColored, FormattedDateRange } from '../_common';
 import ProfitRow from './ProfitRow';
-
+import M from '../_common/M';
```
borisyankov: You can add this to line #2 :)

src/signin/SigninCard.js
```diff
@@ -3,6 +3,7 @@ import LogoSpinner from '../_common/LogoSpinner';
 import ErrorMsg from '../_common/ErrorMsg';
 import InputGroup from '../_common/InputGroup';
 import LanguagePicker from '../_common/LanguagePicker';
+import M from '../_common/M';
```
borisyankov: Consolidate all in one line

PR #34, opened 2015-12-04
Move api call from Components to Actions

qingweibinary:


src/settings/SettingsAddress.js
```diff
@@ -58,18 +57,7 @@ export default class SettingsAddress extends React.Component {
...
-					this.actions.updateSettingFields(req);
-				} else {
-					SettingsAddress.handleUpdateError(response);
-				}
-			},
-			response => {
-				SettingsAddress.handleUpdateError(response);
-			}
-		);
+		this.props.actions.attemptUpdateSettings(req, SettingsAddress.handleUpdateError);
```
borisyankov: tryUpdateSettings is more canonical.
I would still suggest it is updateSettings, as usually try/attempt indicates an alternative call exists without it, one throwing, another not.
After all, most calls might fail for some reason, lets not call them attempt?

src/_common/LanguagePicker.js
```diff
@@ -2,6 +2,16 @@ import React from 'react';
...
+const LanguagePicker = ({onChange, current}) => (
+    <SelectGroup options={languages} value={current} onChange={onChange} />
 );
+
+LanguagePicker.propTypes = {
+    current: React.PropTypes.oneOf(languages.map(ln => ln.value)),
+};
+
+LanguagePicker.defaultProps = {
+    current: 'EN',
```
borisyankov: 'selected' seems the proper name for the LanguagePicker property.
On the other hand the persisted value can be 'currentLanguage' or just 'language'.

src/asset-selector/AssetSelectorCard.js
```diff
@@ -18,8 +18,8 @@ const AssetSelectorCard = ({actions, assets, assetSelector, history, workspace})
 		actions.selectAssetSymbolForTrade(newAsset, oldAsset);
 		if (~window.location.search.indexOf('goback')) history.goBack();
 		if (~window.location.search.indexOf('tick')) {
```
borisyankov: This was not ideal approach, kinda temporary hack, just so you know ;)

src/settings/SettingsAddress.js
qingweibinary: sure ~ was trying to have some name for things that I expect it to fail ~ 
most actions should not fail, so how should we deal with those that might fail?

AFAIK, actions does not return meaningful result by default, which makes it harder to use for read purpose api call 

any thoughts?

src/_common/LanguagePicker.js
qingweibinary: sure =)

src/settings/SettingsAddress.js
borisyankov: You are thinking about actions incorrectly. They have too broad name, but narrower purpose.
An action is a piece of data that changes the state.
It should not fail, the action itself.
And should not return data.
LiveData is handling the requesting and receiving in two places, and it seems to not matter, for most or all cases. If you get a tick, you store it, and don't care who actually requested it.

qingweibinary: Yea, actions should not return data, at least from it's name it should change state, but not returning any of them.

So are you suggesting we keep the api call in components?

PR #41, opened 2015-12-10
Qingwei/refactor store

qingweibinary:


src/_constants/ActionTypes.js
```diff
@@ -53,3 +53,7 @@ export const UPDATE_VIDEO_LIST = 'UPDATE_VIDEO_LIST';
 
 export const WATCHLIST_FAVOR_ASSET = 'WATCHLIST_FAVOR_ASSET';
 export const WATCHLIST_UNFAVOR_ASSET = 'WATCHLIST_UNFAVOR_ASSET';
+
+export const ADD_TOKEN = 'ADD_TOKEN';
```
borisyankov: Just one for 'UPDATE_TOKEN' ? Actions don't need to be too granular.
Now looking at favor and unfavor it could be one action too.

Don't take my word for it though, search for best practices and let's discuss.

src/_reducers/SigninReducers.js
```diff
@@ -7,7 +7,6 @@ import {
 } from '../_constants/ActionTypes';
 
 const initialState = fromJS({
-    token: '',
     email: '',
     password: '',
     language: 'EN',
```
borisyankov: Language too should go.
Or may be we keep token and language, but they are the viewModel thingies, that we do not access anywhere else.

src/_store/configureStore.js
```diff
@@ -6,7 +6,9 @@ import createEngine from 'redux-storage/engines/localStorage';
 import * as ActionTypes from '../_constants/ActionTypes';
 
 const actionsToCache = [
-    ActionTypes.SIGNIN_FIELD_UPDATE,
+    ActionTypes.ADD_TOKEN,
+    ActionTypes.REMOVE_TOKEN,
+    ActionTypes.SERVER_AUTH_FAILED,
     ActionTypes.SERVER_DATA_ASSET_INDEX,
```
borisyankov: SIGNIN_FIELD_UPDATE might still be necessary.

src/_actions/AccountActions.js
```diff
@@ -14,3 +14,17 @@ export const serverDataPayoutCurrencies = serverResponse => ({
     type: types.SERVER_DATA_PAYOUT_CURRENCIES,
     serverResponse,
 });
+
+export const serverAuthFailed = serverResponse => ({
```
borisyankov: Do we need this? We already handle server response for auth, and it can be a success or failure?

src/_constants/ActionTypes.js
qingweibinary: I thought about this, I thought that this could be clearer from the point of caller, not sure if there's best practice for this as it's too use-case specific ~ 

I'll search and reply =)

borisyankov: For these more 'deep' questions you can post on StackOverflow and ping dan_abramov on Twitter to answer. He said this is a good approach ;)

src/_actions/AccountActions.js
qingweibinary: It's my personal preference, I prefer actions to be more specific, instead of having most actions looks like generic setters, so when looking at caller I'll know what's going on directly.

anyway, I do not have strong opinion on it, so I'll change it?

src/_store/configureStore.js
qingweibinary: Why? we are not caching signin as a whole, which should we trigger localstorage save when signin_field_update happens?

src/_constants/ActionTypes.js
qingweibinary: OK, SO had this haha

http://stackoverflow.com/questions/33637740/should-i-use-one-or-several-action-types-to-represent-this-async-action

they generally recommend to combine, still it depends sometime we might have 2 actions that do the same thing now, but diverge later

generic actions never goes wrong though, it will then depends on caller when somethings need to change

anyway for this case, we should combine them.

src/_store/configureStore.js
```diff
@@ -6,13 +6,14 @@ import createEngine from 'redux-storage/engines/localStorage';
 import * as ActionTypes from '../_constants/ActionTypes';
 
 const actionsToCache = [
-    ActionTypes.SIGNIN_FIELD_UPDATE,
```
qingweibinary: @borisyankov 
should we keep it? after clearing this I think this can be merge

borisyankov: Yeah, probably we don't need it.

PR #46, opened 2015-12-13
Trading times filter by date picker

qingweibinary:


src/_actions/WorkspaceActions.js
```diff
@@ -22,10 +22,15 @@ export const updateTickTradeSubmarket = submarket => ({
     submarket,
 });
 
-export const updateTickTradeDate = date => ({
-    type: types.UPDATE_TRADING_TIMES_DATE,
-    date,
-});
+export const updateTradingTimesDate = date => {
```
borisyankov: A request for updating the trade times does not need to be an action, right?
It is not changing the state in itself, the incoming data from server is calling the update for the state.

qingweibinary: Did not notice that, just take a look, that action is react on server data and update trading times data, but `UPDATE_TRADING_TIMES_DATE` is to preserve state of selected date, so that it does not jump back to default when user change language.

maybe the name is confusing ?

borisyankov: Ah, read it again. Makes more sense, now ;)

PR #47, opened 2015-12-14
Payment agent page showing minimal information

qingweibinary:


src/payment-agents/PaymentAgentsItem.js
```diff
@@ -0,0 +1,33 @@
...
+                </div>
+                <div>
+                    <M m="Withdrawal commision: " />
+                    <NumberPlain
+                        currency={paymentAgent.currencies}
+                        value={paymentAgent.withdrawal_commission} />
+                </div>
+            </td>
+            <td>
+                Available banks: ...
```
qingweibinary: I think we should show those banks over here, what do you think?

src/_data/LiveData.js
```diff
@@ -68,6 +69,7 @@ const initAuthorized = (authData, store) => {
     api.getPayoutCurrencies();
     api.subscribeToBalance();           // some call might fail due to backend overload
     api.subscribeToAllOpenContracts();
+    api.getPaymentAgentsForCountry('id');
```
qingweibinary: This is hardcoded just for testing, backend have not support country_code

We should merge after that

borisyankov: Are they translating the country on Account Settings call? If not we can reverse from there.
If they do, need to convert it to country code instead.

src/payment-agents/PaymentAgentsItem.js
borisyankov: We'll show all the info somehow, but we'll consider a better design than what we have for binary.com
For mobile - possibly a full page with the info, but nothing for now.

src/_reducers/PaymentAgentReducers.js
```diff
@@ -0,0 +1,21 @@
+import { Map, List } from 'immutable';
+import { SERVER_DATA_PAYMENT_AGENTS, CHANGE_PAYMENT_AGENT_COUNTRY } from '../_constants/ActionTypes';
+
+const initialState = new Map({
```
borisyankov: I prefer fromJS if nothing fancy is needed.

src/_utils/DataStructureUtils.js
```diff
@@ -0,0 +1,8 @@
+export const twoWayMap = map => {
+    const twoWayMap = map;
+    Object.keys(map).forEach(k => {
```
borisyankov: We'll be using 'reselect' for these types of conversions.
I also don't like our internal structure relying that much on what the API happens to return this day.

src/payment-agents/PaymentAgentsList.js
```diff
@@ -0,0 +1,26 @@
...
+            <table>
+                <thead>
+                    <tr>
+                        <th>Payment Agent</th>
+                        <th>Commissions</th>
+                        <th>Banks</th>
+                    </tr>
+                </thead>
+                <tbody>
+                    { paymentAgents.map(pa => <PaymentAgentsItem key={pa.name} paymentAgent={pa.toJS()}/>) }
```
borisyankov: You better call paymentAgents.toJS() instead of on each item.
But we should do that conversion via selectors (or reselect).

src/_data/LiveData.js
qingweibinary: that was my plan, but it's translated

src/_utils/DataStructureUtils.js
qingweibinary: I'll delete this file, not gonna use it at the moment

src/payment-agents/PaymentAgentsList.js
qingweibinary: sure ~ I'll check on reselect

src/_reducers/PaymentAgentReducers.js
qingweibinary: ok, will change

src/_data/LiveData.js
borisyankov: Then, this really needs fixing ;)

qingweibinary: Raunak says it can be done in one or 2 days, so I think we can wait ?

borisyankov: Sure thing.

PR #48, opened 2015-12-14
Qingwei/virtual account

qingweibinary:


src/_reducers/SignupReducers.js
```diff
@@ -0,0 +1,28 @@
...
+
+export default (state = initialState, action) => {
+    switch (action.type) {
+        case SIGNUP_START: {
+            return state.set('progress', true);
+        }
+        case SIGNUP_FIELD_UPDATE: {
+            return state.set(action.fieldName, action.fieldValue);
+        }
+        case SIGNUP_FAILED: {
```
borisyankov: Let's call it SIGNUP_ERROR
For better diff of FIELD and FAILED. 
(also the passive form of 'failed' seem a bit not what we would call an action, maybe?)

qingweibinary: Sure, will change it, about the noun vs verb, it have been loosely named, but self-explanatory, this remind me of the hardest problem in programming haha

PR #57, opened 2015-12-21
Added resizer

qingweibinary:


qingweibinary:
No ready to merge, wish to discuss with you

src/_common/SizeProvider.js
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+
+export default class SizeProvider extends React.Component {
+    constructor(props) {
+        super(props);
+        this.state = {
+            height: '600px',
```
qingweibinary: this is temporary, should we store this in tradeReducer?

src/_common/SizeProvider.js
```diff
@@ -0,0 +1,39 @@
...
+            width: '1000px',
+        };
+    }
+
+    static propTypes = {
+        windowSizeRelation: React.PropTypes.func.isRequired,
```
qingweibinary: a function that compute width and height with respect of window size, is this what you want?

PR #65, opened 2015-12-22
Add themeProvider

qingweibinary:


src/_common/ThemeProvider.js
```diff
@@ -0,0 +1,15 @@
...
+@connect(state => ({ theme: state.settings.get('theme') }))
+export default class ThemeProvider extends React.Component {
+    static propTypes = {
+        theme: React.PropTypes.string,
+        children: React.PropTypes.object,
+    };
+
+    render() {
+        const theme = this.props.theme;
+        return <div style={{ height: '100%', width: '100%' }} className={(theme === 'dark') ? 'inverse' : ''}>{this.props.children}</div>;
```
borisyankov: Why did you add the 'style' thing?
We will add all via classes.

qingweibinary: without this whole page messed up, I had not figured out the reason, it appear to shrink to width of content ....

borisyankov: Oh, yeah, you are adding another div wrapper, it needs to be there, but can we do that in a way to replace the root div ;)

PR #71, opened 2015-12-28
Qingwei/portfolio fix

qingweibinary:
Problems 
1. Not sure if I should use NumberColored for those figures, looks a little overdone
2. The open_contract call does not seems to be updated very frequent, I'll check with quants, we might need to rely on browser's date


qingweibinary:
Ready to merge ~

PR #74, opened 2015-12-29
Refactor statement shape'

qingweibinary:


src/_reducers/StatementReducers.js
```diff
@@ -1,17 +1,23 @@
...
 
 export default (state = initialState, action) => {
     switch (action.type) {
         case SERVER_DATA_STATEMENT: {
-            return state.set('transactions', action.serverResponse.statement.transactions);
+            const newTx = {};
+            action.serverResponse.statement
+                .transactions
+                .forEach(tx => {
+                    newTx[tx.transaction_id] = tx;
```
borisyankov: This can be done with 'reduce' without the 'const newTx'.
Want to try it? :)

borisyankov:
Review done. OK to merge, but do the reduce thing.

src/_reducers/StatementReducers.js
qingweibinary: OK, will fix =)

PR #76, opened 2015-12-30
Allow min max for resize

qingweibinary:


src/workspace/WorkspaceCard.js
```diff
@@ -29,14 +42,27 @@ export default ({ actions, workspace }) => (
 					]}
 				/>
 			</div>
-			<Resizer className="resizer-vertical" onResize={e => actions.updateWorkspaceField('leftPanelSize', e.x - 4)} />
+			<Resizer
+				className="resizer-vertical"
+				onResize={e => {
+					const update = actions.updateWorkspaceField.bind(null, 'leftPanelSize');
```
qingweibinary: @borisyankov 
Resizer is actually not doing much other than calling function with mouse movement ~
should we refactor it in a way that Resizer actually does more stuff? so that we have less Repetition?

Another problem is that after panel is hidden, to drag it out, client need to drag over minimum, else the panel will just not move, this can be solve by having state of previous mouse event so that we know if user is expanding or shrinking the panel ~

and it's probably cleaner to store the state in resizer instead of workspace ~

borisyankov: Or we can hide the resizer on snap and use icons to pop it up again.

qingweibinary: Yes, that would be simpler ~

PR #78, opened 2016-01-04
Fix digit options barrier range

qingweibinary:


qingweibinary:
@borisyankov conflict resolved, sorry did not notice

PR #79, opened 2016-01-05
Update statement and portfolio when transactions happen in app

qingweibinary:
Note: expiry does not work properly on statement, 
The problem is backend does not sell contract on expiry, the actual selling will be trigger by portfolio or statement call

to make it work on our app, we would need to have some daemon like function take consistently check for expiry and perform portfolio call, it's kind of short polling on client side  =(

@borisyankov any thoughts?
in addition, we would not fix this problem for virtual account, as backend do not want to waste resources on them, backend would only fix this for real account, probably by this month

Edit: Oh wait ~ it seems there's a way ... testing...

### DIFF
--- a/src/_actions/PortfolioActions.js
+++ b/src/_actions/PortfolioActions.js
@@ -1,4 +1,6 @@
 import * as types from '../_constants/ActionTypes';
+import { forceUpdateAll } from '../_utils/ApiWorkaroundUtils';
+import { nowAsEpoch } from '../_utils/DateUtils';
 
 export const serverDataPortfolio = serverResponse => ({
     type: types.SERVER_DATA_PORTFOLIO,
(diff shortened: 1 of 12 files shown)


src/_actions/PortfolioActions.js
```diff
@@ -18,6 +18,7 @@ export const serverDataProposalOpenContract = serverResponse => ({
 
 export const updateNow = () => ({
     type: types.UPDATE_NOW,
+    now: Math.floor(Date.now() / 1000),
```
borisyankov: Should that be an utility function? nowAsEpoch()

src/_utils/ApiWorkaroundUtils.js
```diff
@@ -0,0 +1,16 @@
+import * as LiveData from '../_data/LiveData';
+
+export const forceStatementUpdate = () => {
+    const todayEpoch = Math.floor(Date.now() / 1000) - 86400;
```
borisyankov: Looks like this also belongs to DateUtils

src/_actions/PortfolioActions.js
qingweibinary: Yes, this should be

src/_utils/ApiWorkaroundUtils.js
qingweibinary: Not so sure, as it's difficult to name, it actually means the epoch on the starting point of today ~ 
Should I?

borisyankov: Yep, serves as form of a 'comment'.

qingweibinary: Fixed

src/_actions/PortfolioActions.js
qingweibinary: Fixed

PR #83, opened 2016-01-06
Api auth is now handle before Root rendered

qingweibinary:
Changed the way Auth is handle ~ 

following advice from https://github.com/rackt/react-router/issues/1389

### DIFF
--- a/src/_data/Auth.js
+++ b/src/_data/Auth.js
@@ -1,56 +1,52 @@
-import { rehydratedStorePromise, store } from '../_store/configureStore';
+import { store } from '../_store/configureStore';
 import * as LiveData from './LiveData';
 import { updateToken, signinFieldUpdate, updateAppInfo } from '../_actions';
 import { trackUserId } from '../_utils/Analytics';
 
-let isAuthorized = false;
+export const tryAuth = (st) => {
+    const newState = st.getState();
+    if (!newState.account) {
+        return Promise.reject();
+    }
+
+    const token = newState.account.get('token');
+    if (!token) {
+        st.dispatch(signinFieldUpdate('progress', false));
+        st.dispatch(signinFieldUpdate('tokenNotEntered', true));
+        return Promise.reject();
+    }
+    st.dispatch(updateAppInfo('authorized', false));
+    return LiveData.api.authorize(token).then(
+        response => {
+            st.dispatch(signinFieldUpdate('credentialsInvalid', false));
+            trackUserId(response.authorize.loginid);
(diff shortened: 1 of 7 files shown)


src/_data/Auth.js
```diff
@@ -1,56 +1,52 @@
...
+    const token = newState.account.get('token');
+    if (!token) {
+        st.dispatch(signinFieldUpdate('progress', false));
+        st.dispatch(signinFieldUpdate('tokenNotEntered', true));
+        return Promise.reject();
+    }
+    st.dispatch(updateAppInfo('authorized', false));
+    return LiveData.api.authorize(token).then(
+        response => {
+            st.dispatch(signinFieldUpdate('credentialsInvalid', false));
```
borisyankov: Let's use store everywhere instead of 'st' :)
Also you can use bindActionCreators(AllActions, store.dispatch) to then call them by name and not dispatch, seems a bit cleaner.

src/index.js
```diff
@@ -1,5 +1,13 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
+import { rehydratedStorePromise } from './_store/configureStore';
+import * as LiveData from './_data/LiveData';
+import { tryAuth } from './_data/Auth';
 import Root from './_store/root';
 
-ReactDOM.render(<Root/>, document.getElementById('root'));
+rehydratedStorePromise.then(st => {
```
borisyankov: It's in code like this I wish we had good test coverage.  No more comments, is ok.  ;)

borisyankov:
Done with review. Issues are minor, if you want fix, but I can merge too.

qingweibinary:
I will fix them ;)

src/index.js
qingweibinary: What do you mean ?

borisyankov: That is something goes wrong with the call, whole app is down. Just an observation ;)

src/_data/Auth.js
qingweibinary: store cannot be used as `store` is declared on import

src/index.js
qingweibinary: yes ~ it will show that forever spinning logo ~
how we want to present if things went offline ?

borisyankov: Just a note, 'No connection' or 'App offline' would be enough.
I think we need a general exception handler or something to be able to show an error if the app does not load at all.

qingweibinary: @borisyankov ok, could you merge this 1st, then I'll have another branch to handle this ~

PR #85, opened 2016-01-07
Fixed with error msg and auto correct

qingweibinary:
Fix is ugly from UI, but I am not sure how much effort should we put?
Kindly comment ~ @borisyankov 

screen shot 2016-01-07 at 2 18 15 pm


qingweibinary:
Sorry, accidentally rebase ~
By the way, I did not refactor InputGroup to support the min max restriction out of the box, as it does not compatible with current interface, since it's not broken, I think we can leave it

qingweibinary:
Can we have this merged?

PR #86, opened 2016-01-07
Defined business data state

qingweibinary:
@borisyankov 

This is a proposal on how our store should eventually evolves to ~ and also serves as a record ;)
Let's discuss on this PR ?

I plan to split things into UI state and business data ~ what do you think ~

### DIFF
--- a/states-model.js
+++ b/states-model.js
@@ -1,0 +1,213 @@
+/*
+	< > indicate placeholder
+	data are normalized, and thus have deeper nesting
+*/
+
+states: {
+	business: {
+		tick_subscriptions: {
+			<symbol>: {
+				epoch,
+				id,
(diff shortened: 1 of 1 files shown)


borisyankov:
We definitely want to have the properCasing instead of random_naming ;)
We should time this refactor with the introduction of selectors, and possibly RxJS.

qingweibinary:
thought underscore could be a convention for data from store haha
yea, we can plan before introducing new library ~
I guess RxJS should be in binary-live-api, and next-gen will consume it, right ?

borisyankov:
I suspect RxJS will be just in our code, I think it doesn't need to be in binary-live-api, and can be derived either from the promises or even calls. We'll see.

PR #87, opened 2016-01-08
Fix asset selector input for usability

qingweibinary:


public/styles.css
```diff
@@ -450,6 +450,10 @@ select {
     cursor: pointer;
 }
 
+.asset-list tbody tr.focus-like {
+    background-color: rgb(225, 225, 234);
```
qingweibinary: I could not find the style used for focused tr, thus I simply create something ....

src/asset-selector/AssetSelectorItem.js
```diff
@@ -27,14 +27,9 @@ export default class AssetSelectorItem extends React.Component {
...
 		const { asset, onSelect, isFavorite, isSelected } = this.props;
-		const focuser = node => {
-			if (node !== null && isSelected) {
-				node.focus();
-			}
-		};
 
 		return (
-			<tr autoFocus={isSelected} tabIndex={0} ref={focuser} >
+			<tr className={isSelected ? 'focus-like' : ''} tabIndex={0}>
```
qingweibinary: I think we should probably make all tr like this, thoughts?
What we want is probably not autofocus, but appear to be selected

public/styles.css
borisyankov: It should be just .active or tr:focused I think it was. 
Must be in binary-style

src/asset-selector/AssetSelectorItem.js
borisyankov: Naming is hard? focus-like? :)

tr:active or tr:focus works, and is styled.
TabIndex make it selectable.
But I think it bugs some things though.

public/styles.css
qingweibinary: OK, will check ~

src/asset-selector/AssetSelectorItem.js
qingweibinary: tr:focus break input focus ~
when user typed on asset selector query box, having tr:focus will defocus the input box ~
and I notice we just want to show user what is selected, instead of having focus there

qingweibinary:
@borisyankov I renamed the class to `selected` 
and I used the same style from binary.css

I think it's ok to merge if you dont find issue ;)

PR #88, opened 2016-01-08
Show loading message, structure in way that adding dynamic informatio…

qingweibinary:
…n when reconnecting is trivial, for example reconnecting times

### DIFF
--- a/src/_store/AppInfoProxy.js
+++ b/src/_store/AppInfoProxy.js
@@ -1,0 +1,20 @@
+import React, { Component, PropTypes } from 'react';
+import { connect } from 'react-redux';
+import LoadingView from '../loading-view/LoadingView';
+
+@connect(state => ({ appInfo: state.appInfo }))
+export default class AppInfoProxy extends Component {
+    static propTypes = {
+        appInfo: PropTypes.object.isRequired,
+        children: PropTypes.object,
+    };
+
+    render() {
+        const { connected } = this.props.appInfo.toJS();
+        const msg = `
+        Connecting...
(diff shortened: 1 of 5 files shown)


src/_store/AppInfoProxy.js
```diff
@@ -0,0 +1,20 @@
+import React, { Component, PropTypes } from 'react';
+import { connect } from 'react-redux';
+import LoadingView from '../loading-view/LoadingView';
+
+@connect(state => ({ appInfo: state.appInfo }))
+export default class AppInfoProxy extends Component {
+    static propTypes = {
```
qingweibinary: Not sure if this is a poor name, suggestion are welcome
We can later tap into this component to provide additional information if we necessary ~

I kinda over-engineered it as I was thinking to show how many time they failed, and possibly ask them to check their connection if failure exceeded some threshold, but then found out ws failure was not handled well from the library, so I'll leave it here at the moment

borisyankov: I don't think it matters how many times did we try to reconnect.
But it matters how long was it till we had a connection to the server.

So:
1. Silent retries.
2. After, say 5 seconds of being disconnected, we tell them they are no longer online (5 sec might be too long, we need to test)

borisyankov:
Looks ok. Good to merge?

src/_store/AppInfoProxy.js
qingweibinary: I think 5 is ok if there's no connection at all ~

qingweibinary:
OK, please merge ~ and create a card for the silent retry with interval ~

PR #89, opened 2016-01-11
Fixed payout selector layout, fix min max handling

qingweibinary:
I use min-width to fix the payout selector problem, which may not be appropriate ~ 
Kindly advise

### DIFF
--- a/public/styles.css
+++ b/public/styles.css
@@ -119,6 +119,7 @@ table {
 
 .mobile-content {
     padding: .5rem;
+    min-width: 350px;
     flex: 1;
     display: flex;
     align-content: stretch;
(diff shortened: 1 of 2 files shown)


public/styles.css
```diff
@@ -119,6 +119,7 @@ table {
 
 .mobile-content {
     padding: .5rem;
+    min-width: 350px;
```
borisyankov: Yep, no good. :)
There are mobile devices with smaller widths and this will break them.

qingweibinary: Yep ~ feel free to fix this and close this PR if you have simple and elegant solution, I do not have simple css solution right now, need to do some study this later

PR #90, opened 2016-01-11
Qingwei/new trade

qingweibinary:
Not able to complete much today ~
1. PayoutSelector is just mimicking the old one, I removed the currency and added a MinMax button (might be redundant) ~
2. Duration is more complex, 
   - do we want to implement start time option? ie. allow user to choose when to start the contract
   - do we want to allow user choose end time? (this should not be hard, just some epoch transformation)


borisyankov:
1. Min/Max - possibly redundant, yeah ;)
What we will do is a bit more complex (not immediately, future) - We need to set the 'step' value of the input to the number of digits... Default is 1, so you up/down by 1, but if the value is 25 the step becomes 10, if the value is 123, the step becomes 100, effectively moving the first number of the value.

2. We will want to improve a lot on what we have in our current trade page. So if you decide to do it, implement it the simplest possible way.

qingweibinary:
@borisyankov I've updated ~

Notes:
1. I skip forward starting options as of now ~
2. url is at `/test`, but it breaks when you try to access directly, to access it, go to `/mobile` then `/test`
3. Now the view is hardcoded to use the 1st element from the `trade` store, I will probably add some logic that create trade data + trade view and wire them up ~

Suggestion: 
1. We need to change tick trade so that all trade view share the same API, all trade data share the same shape, so that data and view are decouple, 1 trade data can be rendered by any kind of trade view or vice versa ~

I think this would play well with your plan of grouping trades by category, each group could generate 1:n of trade component ~

src/_utils/TradeUtils.js
```diff
@@ -20,3 +20,50 @@ export const digitOptions = (from, to) => {
...
 };
+
+export const durationText = unit => {
+    switch (unit) {
+        case 't': return 'Ticks';
+        case 's': return 'Seconds';
+        case 'm': return 'Minutes';
+        case 'h': return 'Hours';
+        case 'd': return 'Days';
+        default: return null;
```
borisyankov: Not needed? If not found return => undefined seems more correct.

src/_utils/TradeUtils.js
```diff
@@ -20,3 +20,50 @@ export const digitOptions = (from, to) => {
...
+    const minUnit = min.slice(-1)[0];
+    const maxUnit = max.slice(-1)[0];
+
+    const minIdx = durationUnit.indexOf(minUnit);
+    const maxIdx = durationUnit.indexOf(maxUnit);
+
+    return durationUnit.slice(minIdx, maxIdx + 1);
+};
+
+export const durationLarger = (a, b) => {
```
borisyankov: These functions look like a good candidate for unit tests ;)
We start cautiously using them, and post-demo - very actively.

src/tick-trade/PayoutSelectorCard.js
```diff
@@ -19,9 +19,9 @@ export default class PayoutSelectorCard extends React.Component {
 	updateValue(val) {
 		const { actions, onChange } = this.props;
 
-		if (val.amount > maxAmount) {
+		if (parseFloat(val.amount) > maxAmount) {
```
borisyankov: +val.amount looks better to me than parseFloat(val.amount) ;)

src/trade/BarrierInput.js
```diff
@@ -0,0 +1,22 @@
+import React, { PropTypes, Component } from 'react';
+import { InputGroup } from '../_common';
+
+export default class BarrierInput extends Component {
+    static propTypes = {
+        name: PropTypes.string.isRequired,
+        onChange: PropTypes.func,
+        defaultValue: PropTypes.number,
+    };
```
borisyankov: I started listing the props ordered alphabetically.
Try it next time, I am also not following that very actively.

src/trade/selectors.js
```diff
@@ -0,0 +1,181 @@
+import { createSelector } from 'reselect';
```
borisyankov: I think the selectors would be next to the reducers themselves?
I've seen this most often, but we can investigate.

src/_utils/TradeUtils.js
qingweibinary: ok ~ should we avoid null at all? I dont really understand the intention of having both

src/tick-trade/PayoutSelectorCard.js
qingweibinary: ok, will fix

src/trade/BarrierInput.js
qingweibinary: OK, good to have convention ~

src/trade/selectors.js
qingweibinary: I actually prefer to group things by feature over layer, so that when we want to change something related to a feature, it's easier to browse in the same directory ~

src/_utils/TradeUtils.js
borisyankov: There are no clear guidelines really about both.
The two shouldn't have existed together really.

In JSX null is used often, but in pure code, I find it is very rarely needed, and since undefined exists anyways, I try not to use it.

src/trade/selectors.js
```diff
@@ -0,0 +1,190 @@
...
+const selectedAssetTicksSelector = state => {
+    const fullTrade = state.trade.get('1').toJS();      // hardcoded 1 as fulltrade
+    const selectedSymbol = fullTrade.symbol;
+    const ticks = state.ticks.get(selectedSymbol).toJS();
+    return {
+        ticks,
+    };
+};
+
+const durationInfo = (opts, duration, unit) => {
```
borisyankov: I think you are overusing the 'Info' term ;)

qingweibinary: yes, I'll try to come out with some better naming ~ naming is hard ;)

borisyankov:
Ready to merge, fix merge conflicts ;)

qingweibinary:
I think it's ok to merge ~

PR #91, opened 2016-01-18
Qingwei/trade rewrite

qingweibinary:


src/_actions/TradeActions.js
```diff
@@ -91,7 +95,7 @@ export const updateTradeParams = (id, fieldName, fieldValue) => ({
 
 export const updatePriceProposalSubscription = tradeID => {
     return (dispatch, getState) => {
-        const tradeObj = getState().trade.get(tradeID).toJS();
+        const tradeObj = getState().trades.get(tradeID).toJS();
```
borisyankov: I feel an action should not depend on the state. Thoughts?

src/_actions/TradeActions.js
```diff
@@ -131,8 +136,22 @@ export const updatePriceProposalSubscription = tradeID => {
             stop_profit: stopProfit,
             stop_loss: stopLoss,
         }).then(response => {
-            dispatch(updateTradeParams(tradeID, 'priceProposalID', response.proposal.id));
-            dispatch(updateProposalByID(response.proposal.id, response.proposal));
+            dispatch(updateTradeParams(tradeID, 'proposalError', undefined));
```
borisyankov: We should investigate if putting too much logic in actions is an anti-pattern.
I think it is, but might be wrong.

src/_utils/ArrayUtils.js
```diff
@@ -1 +1,13 @@
 export const arrayHas = (arr, v) => arr.indexOf(v) > -1;
+
+export const groupByKey = (arr, key) => {
+    const result = {};
```
borisyankov: This is a reducer waiting to happen, in fact I think I've written it somewhere already ;)

src/fulltrade/BarrierCard.js
```diff
@@ -0,0 +1,38 @@
+import React, { PropTypes, Component } from 'react';
+import BarrierInput from './BarrierInput';
+
+export default class BarrierCard extends Component {
+    static propTypes = {
+        barrier: PropTypes.number,
+        barrier2: PropTypes.number,
+        barrier1Info: PropTypes.shape({
```
borisyankov: Info :D

src/_actions/TradeActions.js
qingweibinary: Yes, we should discuss ~ the asynchronous call should stay somewhere, anyway ~

src/fulltrade/BarrierCard.js
qingweibinary: To differentiate them from the actual barrier value ~ any suggestion?

the barrierInfo describe name and default value of barrier1

borisyankov: Difference: barrier vs barrierInfo.value ?

qingweibinary: barrier is the actual form params to be used for price proposal ~
barrierInfo.value is some value suggested from backend, I'll name it defaultValue since I confuse you

borisyankov: Then it can be:
barrier {
 defaultValue,
 value,
 name
}

PR #93, opened 2016-01-19
Qingwei/remove trade button

qingweibinary:
this PR already merged 
with 
https://github.com/binary-com/binary-next-gen/pull/92

I'll close 
https://github.com/binary-com/binary-next-gen/pull/92


src/fulltrade/TradePanel.js
```diff
@@ -141,6 +142,11 @@ export default class TradePanel extends Component {
         actions.purchaseByTradeID(id);
     }
 
+    onClosePanel() {
+        const { actions, id } = this.props;
```
qingweibinary: @borisyankov here is the button to remove tradepanel ;)

src/fulltrade/TradesContainer.js
```diff
@@ -15,8 +15,11 @@ export default class TradesContainer extends React.Component {
     };
     createTrade() {
         const { trades, actions } = this.props;
```
qingweibinary: @borisyankov here is the button to create additional trade panel ~

The layout probably should be in this component ~ I've added some logic here, if we want to keep container as pure data layer, we can refactor

PR #96, opened 2016-01-21
Qingwei/complete full trade

qingweibinary:
OK to merge ;)


src/_reducers/TradeReducer.js
```diff
@@ -37,7 +37,10 @@ export default (state = initialState, action) => {
             return initialState;
         }
         case DESTROY_TRADE: {
-            return state.delete(action.id);
+            if (state.size > 1) {
```
qingweibinary: Thought about your comment on actionsCreator should not depends on state ~
I think actionsCreator can depends on state when deciding about side-effects, otherwise, all mutation logic should go into reducer, since reducer have access to state and is easy to test ~

borisyankov: Not sure I ever said that ;)
Definitely a reducer should not depend on api or outside thing, nor should call api or outside service.

PR #97, opened 2016-01-22
Fix react binding problem with radio input

qingweibinary:
@borisyankov I did some dirty trick to make it work, I mimic RadioGroup with buttons ~ and copy the styling ~

I think it is acceptable now ;)
A real fix might require React's team help, I've submitted an issue but have yet provide them test case, will do it when we're less occupied

### DIFF
--- a/public/styles.css
+++ b/public/styles.css
@@ -582,3 +582,22 @@ select {
     font-size: 1.5rem;
     line-height: 2 !important;
 }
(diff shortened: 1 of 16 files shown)


qingweibinary:
Do not merge, something's broken =(

qingweibinary:
OK to merge now ~

PR #98, opened 2016-01-22
Nuru/trans

DanSallau:
Hi,

Attached contained the untranslated text on next-gen.


borisyankov:
You should have either pullled latest or rebased.
See the 'Files' tab, should be shorter ;)

PR #103, opened 2016-01-26
Qingwei/unit test

qingweibinary:


package.json
```diff
@@ -72,6 +73,7 @@
     "mocha": "^2.3.4",
     "react-addons-perf": "^0.14.6",
     "react-addons-test-utils": "^0.14.6",
+    "redux-mock-store": "0.0.6",
```
qingweibinary: @borisyankov not sure if we want to have this ~ 
any thoughts?

useful to test smart component =)

PR #105, opened 2016-01-27
Qingwei/unit test

qingweibinary:


qingweibinary:
@borisyankov it's not supposed to be merge ;)

I somehow remove dependency of json-loader using gulp, if u think it's fine I'll submit a proper fix tomorrow

another thing is the webpack UglifyJSPlugin, we might need to find an alternative ~ there's an open issue on webpack
https://github.com/webpack/webpack/issues/537

I did not finish reading as the options from the thread fix the issue ~

borisyankov:
While we are on the topic, you should try out webpack 2 (it is in beta but soon to be released)
It has tree shaking to optimize unused code.

qingweibinary:
@borisyankov good to merge ~

PR #109, opened 2016-01-29
Merge portfolio related state into single store

qingweibinary:


src/_selectors/FullTradeSelectors.js
```diff
@@ -126,17 +126,40 @@ const durationSecHelper = duration => {
     return durationToSecs(d, u);
 };
 
+const shouldBlockExist = (min, max, unit) => {
```
borisyankov: Not a good name, no idea what this means ;)
Also we should extract to more general biz logic module, this is not selector specific.

qingweibinary: I suggest we do the extraction later ~
I'll move this function to private function and rename it so that full trade works better

PR #110, opened 2016-01-29
Nuru/unit test

DanSallau:
Hi,

Attached contained few unit_test and the istanbul coverage change for your perusal . 

Regard;


borisyankov:
Tests are not passing.

package.json
```diff
@@ -7,7 +7,7 @@
     "test:eslint": "webpack --config webpack.config.dev.js",
     "test:unit": "mocha --compilers js:babel-core/register ./src/**/__tests__/*.js",
     "test:watch": "npm test -- --watch",
-    "test:coverage": "babel-node ./node_modules/istanbul/lib/cli cover ./node_modules/mocha/bin/_mocha ./src/_common/__tests__/*.js",
+    "test:coverage": "babel-node ./node_modules/.bin/babel-istanbul cover ./node_modules/.bin/_mocha ./src/**/__tests__/*.js",
```
qingweibinary: I think we do not want to the _common

src/_common/__tests__/Clock-test.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import expect from 'expect';
+import { shallow,render } from 'enzyme';
+import { FormattedTime } from 'react-intl';
```
qingweibinary: I think we should remove unused import

src/_common/__tests__/Clock-test.js
```diff
@@ -0,0 +1,19 @@
...
+import Clock from '../Clock';
+
+describe('Clock',()=>{
+	const wrapper = shallow(<Clock time={ new Date()}> </Clock>);
+    const time = new Date(wrapper.node.props.value);
+	it("The day is Today's day ",()=>{
```
qingweibinary: The `it` keyword actually refer to the object you want to test, so it's suppose to form a sentence like

`it('should show day as Today's day') `

src/_common/__tests__/Collapsible-test.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import expect from 'expect';
+import {createRenderer} from 'react-addons-test-utils';
+import Collapsible  from '../Collapsible';
+
+describe('Collapsible', ()=>{
+	it('works', ()=>{
```
qingweibinary: more description, it would help new contributor understand what the component does

src/_common/__tests__/DesktopPage-test.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import expect from 'expect';
+import { shallow,render } from 'enzyme';
+//import configureStore from 'redux-mock-store';
```
qingweibinary: Forgot to uncomment ?

src/_utils/__tests__/ArrayUtils-test.js
```diff
@@ -1,7 +1,7 @@
 import expect from 'expect';
-import { groupByKey } from '../ArrayUtils';
+import { groupByKey,objectToArray } from '../ArrayUtils';
 
-describe('groupByKey', () => {
+describe('ArrayUtils', () => {
```
qingweibinary: Make it consistent, if the test is only about groupByKey, then `describe('groupByKey')`

src/_utils/__tests__/DateUtils-test.js
```diff
@@ -18,3 +18,57 @@ describe('epochToDateString', () => {
         expect(str).toEqual('2001-09-09');
     });
 });
+describe('secondsToTimeString', () => {
+    it('convert seconds to minutes',() =>{
```
qingweibinary: I think this is misleading, AFAIK, it convert seconds into HH:mm

qingweibinary: I think it's redundant, as it's covered in the test case below

src/_utils/__tests__/DateUtils-test.js
```diff
@@ -18,3 +18,57 @@ describe('epochToDateString', () => {
...
+
+describe('dateToEpoch', () => {
+    it('convert date to epoch equivalent', ()=>{
+        expect(dateToEpoch(new Date())).toEqual(Math.floor(new Date()/1000));
+    });
+});
+
+describe('timeStringSmaller', () => {
+    it('Comapare two time strings and determine if the first is smaller', ()=>{
+        expect(timeStringSmaller('010203','020203')).toEqual(true);
```
qingweibinary: I think your input format is wrong ~ it should be `HH:mm:ss`
Maybe we should do some checking with the input format though, can you fix it?

qingweibinary:
@nuruddeensalihu 
You need editor to support eslint, as some of your syntax does not follow our eslint rules ~

src/_common/__tests__/Clock-test.js
borisyankov: Agree, think a lot on the description, it is very imortant.
Also, in general the now time is not a good idea to depend on something that 'ticks'. In rare cases possible for some time to have passed and the times to differ.
Also consider running the const wrapper and const time inside the it() block. Tests should be very isolated from each other.

src/_common/__tests__/Collapsible-test.js
borisyankov: Agree. That part of the test is at least half of the task ;)

src/_utils/__tests__/DateUtils-test.js
borisyankov: You can also have several tests for one thing.
Test edge cases, that is important, and test the negative.
For example reverse the numbers and do a toEqual(false)

Very important to spend more time on the descrition.
For example in this case it might be:
'should compare two time strings'

On another note, a comparer function usually returns a -1, 0 or 1, not sure I like this function ;)

qingweibinary:
@nuruddeensalihu there's merge conflict ;)

borisyankov:
Are you working from different forks?
If you are not, that is why may be there are more commits than necessary?

We should do a rebase for this PR, which will make all commits into a single commit.

PR #114, opened 2016-02-01
Qingwei/fix trade ui

qingweibinary:
This depends on 
https://github.com/binary-com/binary-next-gen/pull/109

### DIFF
--- a/src/_selectors/FullTradeSelectors.js
+++ b/src/_selectors/FullTradeSelectors.js
@@ -5,9 +5,8 @@ import { durationToSecs } from '../_utils/TradeUtils';
 import {
     epochToUTCTimeString,
     nowAsEpoch,
-    timeStringBigger,
-    timeStringSmaller,
     splitSecsToUnits,
+    timeStringIsBetween,
 } from '../_utils/DateUtils';
 import { marketTreeSelector } from './AssetSelectors';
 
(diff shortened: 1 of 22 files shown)


src/_selectors/FullTradeSelectors.js
```diff
@@ -255,15 +255,24 @@ const availableAssetsFilter = (assets, times, now) => {
...
             availabilities[s.symbol] = true;
-        } else if (timeStringBigger(nowInTimeString, open) && timeStringSmaller(nowInTimeString, close)) {
-            availabilities[s.symbol] = true;
+        } else if (open.length === 1) {
+            if (timeStringBigger(nowInTimeString, open[0]) && timeStringSmaller(nowInTimeString, close[0])) {
+                availabilities[s.symbol] = true;
+            }
+        } else if (open.length === 2) {
+            if (
+                (timeStringBigger(nowInTimeString, open[0]) && timeStringSmaller(nowInTimeString, close[0])) ||
```
borisyankov: Why are we using string actually?
Because of input type=date ?
If it was Date it would just be a '>' comparison.

src/_selectors/FullTradeSelectors.js
```diff
@@ -255,15 +255,24 @@ const availableAssetsFilter = (assets, times, now) => {
...
             return;
         }
-        const open = s.times.open[0];
-        const close = s.times.close[0];
+        const open = s.times.open;
+        const close = s.times.close;
 
         // Assuming closing time is larger than open time
         // TODO: some of Random does have time which does not fit into assumption
         if (s.name.indexOf('Random') > -1) {
```
borisyankov: Will this work for a different language?
Also str.indexOf(...) > -1 can be str.includes(..)

src/_utils/ArrayUtils.js
```diff
@@ -14,3 +14,11 @@ export const objectToArray = obj =>
     typeof obj.toArray !== 'undefined' ?
         obj.toArray() :
         Object.keys(obj).map(key => obj[key]);
+
+export const arrayEqual = (a, b) => {
+    if (a.length !== b.length) {
+        return false;
+    }
+
+    return a.filter((x, idx) => x !== b[idx]).length === 0;
```
borisyankov: use [].some() function instead of filter()

borisyankov:
Done with review. Fix small issues and merge conflict and I merge ;)

src/_selectors/FullTradeSelectors.js
qingweibinary: AFAIK symbol name are never translated, nonetheless we should not rely on this, I am fixing this issue as there's other exceptions

src/_selectors/FullTradeSelectors.js
qingweibinary: Because that's what backend returns ~
And it represent time instead of Date, we need to use string like
'09:09:00' and '12:00:00' to determine if the market is close or open, any idea of improvement ?

borisyankov: Definitely we will have a transform on the server_data inputs. We can even have it as an extension to live-api. Not today, but lets consider it soon.

src/_selectors/FullTradeSelectors.js
borisyankov: You can try testing in Chinese, if everything still works.

qingweibinary: Removed =)

src/_selectors/FullTradeSelectors.js
qingweibinary: Good idea about extending live-api ~
Yea, right now let's have it as string :(

PR #118, opened 2016-02-02
Disabled trade panel when data have not reached

qingweibinary:


src/_actions/TradingOptionsActions.js
```diff
@@ -7,12 +7,17 @@ export const updateTradingOptions = (symbol, opts) => ({
     opts,
 });
 
-export const getTradingOptions = symbol =>
+export const getTradingOptions = (symbol, cb) =>
```
borisyankov: cb = callback, or something like onDone, cb => bad name ;)
Do this as parameter cb = () => {}
Then you will not need to check for existing,

borisyankov:
I merged the PR manually with everything but the last commit. Pull latest.

qingweibinary:
@borisyankov the last commit is the intraday definition ??

PR #121, opened 2016-02-02
nuru/my_test

DanSallau:
Hi,

Attached contain few unit test for _actions for your perusal.


src/_actions/__tests__/AppConfigActions-test.js
```diff
@@ -0,0 +1,14 @@
...
+import expect from 'expect';
+
+describe('AppConfigActions',()=>{
+  it('should be equal to the expectedAction',()=>{
+      const expectedAction = {
+        type: UPDATE_APP_CONFIG,
+        field: 'balance',
+        value: 3,
+      };
+      expect(updateAppConfig('balance',3)).toEqual(expectedAction);
```
qingweibinary: @borisyankov @nuruddeensalihu do you think we can skip these test, seems not very valuable ~

I think testing on action creators is more worth it, as there's logic inside

Thoughts?

borisyankov: @nuruddeensalihu these are actually action creators.
Notice they are actually functions that return the action object.
I'd say the no logic ones will not fail.
There are others that can fail, they have api calls, but on the other hand they are hard to test.

borisyankov:
Please examine how to run ESLint on the tests files too.
This will both help with the formatting these tests and in future tests.

PR #122, opened 2016-02-03
Qingwei/fix trade ui

qingweibinary:


src/charting/EChart.js
```diff
@@ -25,6 +25,14 @@ export default class EChart extends React.Component {
         this.echart.setOption(nextProps.options);
     }
 
+    componentDidUpdate(nextProps) {
+        if (nextProps.style !== this.props.style) {
+            this.echart.resize();
+        }
+
+        this.echart.setOption(nextProps.options);
```
borisyankov: These setOption calls are updating the component. We definitely need to test if we are not calling them too frequently.

src/fulltrade/TradePanel.js
```diff
@@ -51,7 +51,7 @@ const createDefaultBarriers = (contracts, category, type, duration, durationUnit
     let expiryType;
     if (durationUnit === 't') {
         expiryType = 'tick';
-    } else if (durationToSecs(duration, durationUnit) <= 86400) {
+    } else if (durationToSecs(duration, durationUnit) < 86400) {
```
borisyankov: Use 60 * 60 * 24 to make the constant more obvious

src/fulltrade/TradePanel.js
```diff
@@ -358,12 +373,13 @@ export default class TradePanel extends Component {
         const receipt = trade.receipt;
         const isTick = trade.durationUnit && trade.durationUnit.slice(-1) === 't';
         const isBelow2Min = isTick || durationToSecs(trade.duration, trade.durationUnit) < 120;
-        const isIntraDay = durationToSecs(trade.duration, trade.durationUnit) <= 86400;
+        const isIntraDay = durationToSecs(trade.duration, trade.durationUnit) < 86400;
```
borisyankov: isIntraday => util func outside of here, TradeUtils / DateUtils?

src/_selectors/AssetSelectors.js
```diff
@@ -3,6 +3,7 @@ import { toPlainJS } from '../_utils/ObjectUtils';
 
 export const assetsSelector = state => toPlainJS(state.assets);
 
+// return a tree structure, hierarchy as [market -> submarket -> symbol]
 export const marketTreeSelector = createSelector(
```
borisyankov: BTW, I am not sure the way I implemented the market tree is ideal, or should have used arrays instead of objects. (just a note :)

src/charting/EChart.js
qingweibinary: Yes, opps, I'll remove changes in this file, 
I was fiddling around to try fix the resizing yesterday, maybe I revert the changes to a wrong commit

src/fulltrade/TradePanel.js
qingweibinary: OK ~

src/fulltrade/TradePanel.js
qingweibinary: This is highly context specific, but ok, since we already have durationToSecs which is also very context specific =P

src/_selectors/AssetIndexSelectors.js
```diff
@@ -23,7 +24,32 @@ export const shownAssetIndexRowsSelector = createSelector(
     }
 );
 
-export const assetIndexHeadersSelector = createSelector(
+const assetIndexRowToDuration = row =>
+    row ? `${row[2]}–${row[3]}` : '—';
+
+const assetIndexTableSelector = createSelector(
+    shownAssetIndexRowsSelector,
```
qingweibinary: One simple improvement we can do is to store the proper structure in redux-store instead of transforming it afterwards ~

Right now I converted them into a tabular form, ie. 2d arrays
I've check and seems only AssetIndex component is depending on this data, so it's probably save to change redux store

The structure is 
```javascript
[
  ['Up/Down', 'Ends In/Out', ...],
  ['Random 100', '5t-10t', ....],
  ['Random 50', '5t-10t', ....],
  ....
]
```

Do you think it's ok to store this 2d arrays in redux store instead?

borisyankov: Yes, I am considering this data transformation layer to be in LiveApi even.
We can add it as LiveDataTransform.js so other can use it too.

borisyankov: It will be optional, but will hook on data received and will fire the events wit the transformed data. 'Data Transform Middleware' :)

borisyankov:
Ready to merge?

qingweibinary:
Ready to merge after test pass

PR #123, opened 2016-02-03
Nuru/reducer test

DanSallau:
Hi,

Attached contained few unit Test for your perusal.


qingweibinary:
@nuruddeensalihu test failed

PR #131, opened 2016-02-15
Fix asset index

qingweibinary:
Please review, hope I did not override any last weeks changes


src/tick-trade/TickTradeContainer.js
```diff
@@ -26,9 +26,9 @@ export default class TickTradeContainer extends Component {
         return (
             <div>
                 {allTickId.map(id => {
-                    const trade = trades[id];
+                    const trade = trades.get(id).toJS();
```
qingweibinary: This could be mistake, let me check

PR #133, opened 2016-02-16
Add support for CashierLock

DanSallau:
Hi,

The attached changes cater for the cashierLock or security settings as sometimes referred to. The cashier lock should work fine now.


src/settings/SettingsSecurity.js
```diff
@@ -1,24 +1,118 @@
...
+
+        const req = {
+            lock_password: state.cashierlockpassword1,
+            unlock_password: state.cashierlockpassword1,
+        };
+
+        const isValid = (pw1, pw2) => ((pw1 === pw2) && (pw1.length > 6));
+        if (!!(settings.cashier_password)) {
+            delete req.lock_password;
+            LiveData.api.setCashierLock(req)
```
borisyankov: Use async/await.

src/settings/SettingsSecurity.js
```diff
@@ -1,24 +1,118 @@
...
+        this.setState({
+            cashierlockpassword1: '',
+            cashierlockpassword2: '',
+        });
+    }
+
+	onClick() {
+        const state = this.state || {};
+        const { settings } = this.props;
+        const pwd1 = state.cashierlockpassword1;
```
borisyankov: Proper casing would be cashierLockPassword1, but also password1 is enough.

src/settings/SettingsSecurity.js
```diff
@@ -1,24 +1,118 @@
...
+        const pwd2 = state.cashierlockpassword2;
+
+        const req = {
+            lock_password: state.cashierlockpassword1,
+            unlock_password: state.cashierlockpassword1,
+        };
+
+        const isValid = (pw1, pw2) => ((pw1 === pw2) && (pw1.length > 6));
+        if (!!(settings.cashier_password)) {
+            delete req.lock_password;
```
borisyankov: Rarely if ever use delete. Create new objects when needed.

src/settings/SettingsSecurity.js
```diff
@@ -1,24 +1,118 @@
...
+
+    clearState() {
+        this.setState({
+            cashierlockpassword1: '',
+            cashierlockpassword2: '',
+        });
+    }
+
+	onClick() {
+        const state = this.state || {};
```
borisyankov: You should set state in the constructor. Then you definitely do not need to initialize it.

src/settings/SettingsSecurity.js
```diff
@@ -1,24 +1,118 @@
...
+        const pwd1 = state.cashierlockpassword1;
+        const pwd2 = state.cashierlockpassword2;
+
+        const req = {
+            lock_password: state.cashierlockpassword1,
+            unlock_password: state.cashierlockpassword1,
+        };
+
+        const isValid = (pw1, pw2) => ((pw1 === pw2) && (pw1.length > 6));
+        if (!!(settings.cashier_password)) {
```
borisyankov: Extract the two conditions of 'if' into two functions.

PR #138, opened 2016-02-22
Fix News/Daily Report bug, Address the browser compatibility issue

DanSallau:
Hi ,

Attached contained changes that should address our Daily report bugs/issues. The browser compatibility and the current Article issues should be addressed.


borisyankov:
Looks good ;)

PR #140, opened 2016-02-22
Fix verify_code bug

DanSallau:
Hi,

Attached contained a little fix for the verify_code of the create account feature.

### DIFF
--- a/src/create-account/AccountInfoForm.js
+++ b/src/create-account/AccountInfoForm.js
@@ -47,14 +47,14 @@ export default class CreateAccountCard extends Component {
     async proceed() {
         const { actions, email } = this.props;
 
-        await actions.createAccountFieldUpdate('validatedOnce', true);
-        await actions.createAccountFieldUpdate('progress', true);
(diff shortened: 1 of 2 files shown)


src/create-account/AccountInfoForm.js
```diff
@@ -46,12 +46,16 @@ export default class CreateAccountCard extends Component {
...
         const { actions, email } = this.props;
+
         actions.createAccountFieldUpdate('validatedOnce', true);
         actions.createAccountFieldUpdate('progress', true);
         if (this.emailValid() && this.passwordValid() && this.confirmationValid()) {
             LiveData.api.verifyEmail(email, 'account_opening').then(() => {
                 actions.createAccountFieldUpdate('step', 1);
                 actions.createAccountFieldUpdate('progress', false);
+            })
+            .catch((err) => {
```
borisyankov: Use async/await here too, with a try catch block

nuruddeensalihu: Wanted to , but on a second thought i didn't considering the light action performed.

PR #143, opened 2016-02-23
Add support for virtual account topup

DanSallau:
Hi,

Attached contained few changes that would enable Virtual account users to topup their balance once it is below USD1000.


src/settings/SettingsGeneral.js
```diff
@@ -8,14 +12,29 @@ export default class SettingsGeneral extends Component {
...
+		balance: PropTypes.string.isRequired,
+        settings: PropTypes.object,
 	};
 
 	onThemeChange(e) {
 		this.props.actions.updateAppConfig('theme', e.target.value);
 	}
+	topUpClick() {
+        const { actions } = this.props;
+		LiveData.api.topUpVirtualAccount().then(response => {
```
borisyankov: async / await here

src/settings/settingsSelectors.js
```diff
@@ -2,9 +2,11 @@ import { createStructuredSelector } from 'reselect';
 import { appConfigSelector, settingsSelector } from '../_store/directSelectors';
 
 const loginidSelector = state => state.account.get('loginid');
+const balanceSelector = state => state.account.get('balance');
```
borisyankov: Move both of these to directSelector file.

src/_common/VirtualTopUpConfirmation.js
```diff
@@ -0,0 +1,55 @@
...
+                <td><M m="Amount" /></td>
+                <td>
+                    <NumberPlain value={response.topup_virtual.amount} />
+                </td>
+            </tr>
+            </tbody>
+        </table>
+        <br/>
+        <button><M m="Back" /></button>
+    </div>)
```
borisyankov: It seems that the two possibilities are very similar in code, should not duplicate that much html.

PR #145, opened 2016-02-26
Nuru/Fix_Language_Change_Bug

DanSallau:
Hi,

Attached conatined few changes that should address the language change login/authorization issues . The second invalidcontract error might be as a result of an expired contract id and has not yet been addressed by this PR. Would look further for that Monday.


src/_data/LiveData.js
```diff
@@ -45,11 +45,14 @@ const subscribeToWatchlist = st => {
     api.subscribeToTicks(favs.toJS());
 };
 
-export const changeLanguage = ln => {
-    api.changeLanguage(ln);
-    api.getActiveSymbolsFull();
-    api.getAssetIndex();
-    api.getTradingTimes();
+export const changeLanguage = async (dispatch, ln) => {
+    api.events.on('authorize', response => response.error ?
```
borisyankov: We want to use the promise + await approach in this case.
Though I am not sure even if this needs to be here?!?

src/fulltrade/FullTradeCard.js
```diff
@@ -51,6 +51,7 @@ export default class GenericTradeCard extends Component {
         trade: PropTypes.object.isRequired,
         type: PropTypes.oneOf(['tick', 'full']).isRequired,
         tick: PropTypes.array,
+        isAuthorized: PropTypes.bool.isRequired,
```
borisyankov: May be we want to abstract this more like 'isAvailableForTrade'.
Need to discuss it three of us later.

src/tick-trade/TickTradeCard.js
```diff
@@ -66,6 +67,7 @@ export default class TickTradeCard extends Component {
                 />
                 <BuyButton
                     askPrice={askPriceFromProposal(trade.proposal)}
+                    isAuthorized={isAuthorized}
```
borisyankov: Too much 'leak' of information. We don't want each component to be aware of authorization. Disabled should be used in this case.

src/tick-trade/tickTradeSelectors.js
```diff
@@ -37,4 +37,5 @@ export const tickAssetsSelector = createSelector(
 export default createStructuredSelector({
     ...singleTradeSelectors,
     assets: tickAssetsSelector,
+    isAuthorized: state => (state.appState.get('authorized')),
```
borisyankov: Use from directSelectors

src/trades/__tests__/tradeListSelectors-test.js
```diff
@@ -10,6 +10,7 @@ describe('tradeListSelectors', () => {
         trades: fromJS([]),
         tradingOptions: fromJS([]),
         tradingTimes: fromJS([]),
+        appState: fromJS({ authorized: true }),
```
borisyankov: This is indicative of a potential problem I noticed.
A selector depending on too many reducers might not be ideal.
Need discussion between 3 of us.

PR #146, opened 2016-02-29
Replaced chart with binary-charts

qingweibinary:
Do not merge as there's some UI bugs


package.json
```diff
@@ -41,9 +41,10 @@
     ]
   },
   "dependencies": {
+    "binary-charts": "0.0.3-patch2",
     "binary-live-api": "*",
     "classnames": "^2.2.3",
-    "echarts": "^3.0.2",
+    "echarts": "^3.1.1",
```
borisyankov: We probably don't need echarts as a dependency anymore?

qingweibinary: if we replace all existing chart then yes

PR #150, opened 2016-03-07
Nuru/Full_trade_bug

DanSallau:
Hi,

Attached contained few changes to address our full trade issues. The FullTrade page and the tick trade pages should now open as expected. Though there might be some few fixes/bugs that would love to address tomorrow.

Regard;


src/fulltrade/FullTradeCard.js
```diff
@@ -57,11 +57,11 @@ export default class BaseTradeCard extends Component {
                     assetName={selectedSymbol}
                     onClosePanel={() => actions.removeTrade(index)}
                 />
-                {tick && <TradeChart
+                {(tick && Object.keys(tick).length > 0) && <TradeChart
```
qingweibinary: Why is tick now an object instead of array?
I think we can make selector always return array, so that we dont have to check `undefined` here, just length will do, 

In addition, seems @borisyankov suggested to let chart handle empty array, so we could probably just pass it in, and enforce the props type so that it warns when it's undefined

src/fulltrade/FullTradeCard.js
```diff
@@ -57,11 +57,11 @@ export default class BaseTradeCard extends Component {
...
                     symbol={selectedSymbol}
                     history={tick}
                 />}
-                <FullTradeParams {...this.props} />
+                {(contract && Object.keys(contract).length > 1) && <FullTradeParams {...this.props} />}
```
qingweibinary: Is contract ever gonna be an empty object?

nuruddeensalihu: Yes It usually is at the beginning usually before a response is returned i guess.

qingweibinary: If that's the case, how about undefined?
I believe there should never be a empty object though, empty object is `{}` right ?

borisyankov: We are trying to patch a bug by not fixing the bug I suspect.
Two other questions:
* Is contracts === undefined a bug?
* if it isn't possibly FullTradeparams can still handle a null contract by having a defaultValues = { contract: {} };

src/fulltrade/FullTradeCard.js
nuruddeensalihu: Its still an array. The use of Object.keys there is not necessary though not an issue :wink .

borisyankov: Yep, we will have transitioning states for the chart. Visually there should always be a chart on screen even if empty. In fact we might decide to return a <div>Loading...</div> from the chart component (or something similar).

Same thing for trading components, we should rarely or never just hide components (visually look bad). Alternative - disabled or loading states

PR #153, opened 2016-03-08
Fix new trade bug when initial trade asset is different from default

qingweibinary:


src/_actions/TradeActions.js
```diff
@@ -61,10 +61,24 @@ export const setQuickTradeField = (symbol, tradeType, field, value) => ({
     value,
 });
 
-export const createTrade = symbol => ({
-    type: types.CREATE_TRADE,
-    symbol,
-});
+export const createTrade = symbol =>
+    (dispatch, getState) => {
+        const activeIdx = getState().workspace.get('activeTradeIndex');
```
borisyankov: Not liking the super strong dependency of the action to the whole state structure.
Possibly all stuff you get from the store can be passed as parameters?

src/_actions/TradeActions.js
```diff
@@ -61,10 +61,24 @@ export const setQuickTradeField = (symbol, tradeType, field, value) => ({
...
+        const ticksExist = getState().ticks.get(symbolUsed);
+
+        const contractP = !!contractExist ? Promise.resolve() : LiveData.api.getContractsForSymbol(symbolUsed);
+        const ticksP = !!ticksExist ? Promise.resolve() : LiveData.api.subscribeToTick(symbolUsed);
+
+        Promise.all([contractP, ticksP])
+            .then(
+                v => {
+                    console.log(v);
+                    dispatch({ type: types.CREATE_TRADE, symbol: symbolUsed });
```
borisyankov: I am getting what you try to do here, but very skeptical about its robustness.

borisyankov: Looks like redux-saga to me.

src/_actions/TradeActions.js
qingweibinary: Conceptually, createTrade does depends on all these states, I think it's fine, this is the way our UI works ~ =)

definitely possible to pass as params, this would means we shift the dependency to the Component itself instead of actionCreators.

Will do it

src/_actions/TradeActions.js
qingweibinary: Not sure if such coding style are due to coming from classical OOP >.<
The react push and update paradigm makes it difficult to store such logic

As for robustness, I believe it's fine, it will break of course if we change any of those state related

Any suggestion?

PR #157, opened 2016-03-09
Redirect real-account token

qingweibinary:


src/only-virtual/OnlyVirtualPage.js
```diff
@@ -0,0 +1,7 @@
+import React from 'react';
+
+export default class OnlyVirtualPage extends React.Component {
+    render() {
+        return <div>This site currently in Beta version, only Virtual Accounts are allowed.</div>;
```
qingweibinary: This is definitely not good enough, I submitted so you can review, and not sure if we want to merge into master, or only merge to a release branch.

src/_data/LiveData.js
```diff
@@ -81,7 +81,7 @@ const initAuthorized = (authData, store) => {
     api.subscribeToAllOpenContracts();
     subscribeToWatchlist(store);
 
-    const isVirtual = authData.authorize.loginid.startsWith('VRTC');
+    const isVirtual = authData.authorize.is_virtual === '1';
```
borisyankov: Extract logic to module. (simple but reusable)

src/only-virtual/OnlyVirtualPage.js
borisyankov: Just use src/_utils/MessagingUtils.js and showError or showInfo

qingweibinary: Alert?

src/_data/LiveData.js
qingweibinary: Not sure where should I place this, how about let it stay here until we actually find we need them more?

src/only-virtual/OnlyVirtualPage.js
borisyankov: Nope, it is a showError. Currently alert, later whatever we want it to be.
Instead of rebuilding these components with just a div and a text inside them, dozen times...

qingweibinary: OK ~

qingweibinary:
@borisyankov I think this is ok to merge ~

PR #158, opened 2016-03-11
Fix select All bug in asset

DanSallau:
Hi,

Attached contained some few changes that should address the asset picker issue.


borisyankov:
Good find, but I don't think we should solve the bug like this.
Actually it is a bug in React 15:
https://github.com/facebook/react/pull/6228

PR #168, opened 2016-03-18
Fix start date duration setting bug

qingweibinary:


src/fulltrade/FullTradeParams.js
```diff
@@ -26,6 +26,11 @@ import { createDefaultType, createDefaultDuration, createDefaultBarriers } from
  * 4. digit trade is always ticks only
  * 5. barriers are non available for contract below 2 minutes
  * 6. forward starting does not have barriers
+ *
+ *
+ * TODO: reform the whole dependency mess,
+ * possibly a better approach would be notify user for wrong info instead of change to correct value automatically,
```
borisyankov: May be not, may be smarter defaults. We'll see. Refactor though, agree ;)

PR #170, opened 2016-03-18
Make stake, profit and buy button inline

qingweibinary:


src/_common/M.js
```diff
@@ -15,11 +15,13 @@ export default class M extends Component {
 		const { m = '', values } = this.props;
 
 		return (
-			<FormattedMessage
-				id={m}
-				defaultMessage={m}
-				values={values}
-			/>
+			<span {...this.props}>
```
borisyankov: What is the purpose of this?

qingweibinary: To be able to style M, like font-weight etc

borisyankov: That was one of my annoyances that they don't pass a className down to their span, but they discussed it on the Issues part of GitHub and was against it. So this is probably better.

qingweibinary: Yes, I read that and thus select this method, they have an internal id, which is not html id at all

PR #172, opened 2016-03-18
Split updown to risefall and higherlower

qingweibinary:
This changes affects quite a number of things, (code is short though)

I've tried to make sure it is safe, but still we need a careful review, @borisyankov @nuruddeensalihu help me take a look and hope I did not miss anything out

### DIFF
--- a/src/_reducers/TradingOptionsReducer.js
+++ b/src/_reducers/TradingOptionsReducer.js
@@ -3,10 +3,38 @@ import { UPDATE_TRADING_OPTIONS } from '../_constants/ActionTypes';
 
 const initialState = fromJS({});
 
+const splitRiseFallAndHighLow = options =>
+    options.map(opt => {
+        if (opt.contract_category === 'callput') {
+            const cloned = Object.assign({}, opt);
+            switch (opt.barriers) {
+                case 0: {
+                    const enrichedOpt = Object.assign(cloned, {
+                        contract_category_display: 'Rise/Fall',
+                        contract_category: 'risefall',
+                    });
+                    return enrichedOpt;
+                }
+                case 1: {
+                    const enrichedOpt = Object.assign(cloned, {
+                        contract_category_display: 'Higher/Lower',
+                        contract_category: 'higherlower',
+                    });
+                    return enrichedOpt;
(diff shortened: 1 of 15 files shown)


qingweibinary:
This PR is about several things as follow

1. Fix all race condition I've encounter, basically the root cause is that asset picker can change the trade params, thus trade params lose some of the control, and behave incorrectly as the code is develop as a self-contained entity but grow into something more complex.

2. Remove redundant subscription call, the root cause is the same with number 1.

Note: we need to redesign this part (After stable release ;) )

src/_reducers/TradingOptionsReducer.js
```diff
@@ -3,10 +3,38 @@ import { UPDATE_TRADING_OPTIONS } from '../_constants/ActionTypes';
...
+                }
+            }
+        } else return opt;
+    });
+
 export default (state = initialState, action) => {
     switch (action.type) {
         case UPDATE_TRADING_OPTIONS: {
-            return state.set(action.symbol, action.options);
+            const enrichedOptions = splitRiseFallAndHighLow(action.options);
```
borisyankov: Oh man, this naming ;)

src/fulltrade/FullTradeParams.js
```diff
@@ -28,10 +34,13 @@ import { createDefaultType, createDefaultDuration, createDefaultBarriers } from
...
  * 5. barriers are non available for contract below 2 minutes
- * 6. forward starting does not have barriers
+ * 6. For underlying that's under our control, we allow relative and absolute barrier regardless of intraday or not
+ * 7. Relative barrier is always allowed
+ * 8. Forward starting do not have barriers
+ * 9. All underlying have risefall trade
  *
  *
- * TODO: reform the whole dependency mess,
+ * TODO: re-architect the whole dependency mess,
```
borisyankov: 10. All assumptions converted to unit tests :)

qingweibinary: most of this assumption is assumption on data, meaning we are short of metadata, we need to confirm with Quants, and think what to do when assumption breaks

PR #173, opened 2016-03-21
Fix missing locale data bug

DanSallau:
Hi,

Attached contained few changes that should address our locale issue on language change. Should there be any issue with regard to this in the future, would be glad to address.


package.json
```diff
@@ -53,6 +53,9 @@
     "classnames": "^2.2.3",
     "history": "^2.0.1",
     "immutable": "^3.7.6",
+    "lodash.isfunction": "^3.0.8",
+    "lodash.set": "^4.0.0",
+    "lodash.unset": "^4.2.1",
```
borisyankov: These are here by mistake, right?

src/_store/AppConfigProvider.js
```diff
@@ -23,9 +24,11 @@ export default class AppConfigProvider extends Component {
         language: PropTypes.string.isRequired,
         theme: PropTypes.string.isRequired,
     };
-
+    
     render() {
         const { children, language, theme } = this.props;
+        addLocaleData(localeData(locale(language)));
```
qingweibinary: Supposingly render should be pure, is there any justification for this side-effect?

package.json
nuruddeensalihu: Yup, Couldn't run next-gen FRiday until i install the 3 libraries. Was expecting the same thing today . Forgot to check after getting latest. Would get rid of them

src/_store/AppConfigProvider.js
nuruddeensalihu: Is aware of that. Wanted to put it in one of the lifeCycle events. Then i noticed it we are not loading on language change , we are just re-rendering based on on the choice and hence the lifecycle properties are not available on the language change . Our locale data on the other hand need to be loaded on each language change and according to the respective selection. :wink:

qingweibinary: Not entirely sure about `we are not loading on language change`
We can react on language change, either by creating the lifecycle method and call side-effects as needed, or hook on the actions that change language ~

nuruddeensalihu: You are right, there are other events i can explore since beside the ones that are called ones. Thanks for the correction.

borisyankov:
Good job. I like the latest version :)
Please add Japanese language too, and do a new PR with rebase, and it is ready to merge.

PR #178, opened 2016-03-22
Do not show asset when user cannot purchase corresponding contract, w…

qingweibinary:
…hen market close and dont allow start later


qingweibinary:
It throws warning about duplicate keys, i am working on it

PR #179, opened 2016-03-23
Nuru/bug_fix

DanSallau:
Hi,

Attached contained few changes that would address some of the issues reported on our trello. Among them is https://trello.com/c/KO3Jgnz6/7-application-hangs-after-selecting-digit-trades-and-shows-error-in-the-console , https://trello.com/c/7bGcx9pF/4-click-ok-button-in-contract-confirmation-screen-nothing-happens etc.


src/payout-picker/PayoutPickerCard.js
```diff
@@ -40,10 +40,10 @@ export default class PayoutPickerCard extends Component {
 
 	render() {
 		const { trades } = this.props;
-		const { index } = this.props.params;
+		const { id } = this.props.params;
```
borisyankov: It should be index

PR #183, opened 2016-03-25
Fix sell at market validation error

DanSallau:
Hi,

Attached contained few changes that should fix our trade purchase issue for your perusal. There seems to be a little issue with the validation that i failed to noticed yesterday , because i didn't get latest prior to testing. SHould there be any thing with regard to this, please do let me know and would be glad to address. :wink:


src/contract-details/ContractDetailsCard.js
```diff
@@ -61,9 +60,11 @@ export default class ContractDetailsCard extends Component {
 						/>)
 					</button> :
 					<div>
-					{validationError.message ?
-						validationError.message.split(')')[1] :
-						validationError }
+					{(!!validationError) &&
```
borisyankov: You don't need the () and also you don't need to convert it to boolean with the !!.
In fact it is better not to.

PR #185, opened 2016-03-28
Fix hide sell at market button when durationunit is tick

DanSallau:
Hi,

Attached contained few changes that should fix our or hide sell at market button when the duration unit is ticks. Should there be any issue with regard to this, i would be glad to address. Trello https://trello.com/c/0aPKfajT/31-sell-at-market-shouldn-t-be-shown-for-trades-with-duration-as-ticks

### DIFF
--- a/src/barrier-picker/SpreadBarrierCard.js
+++ b/src/barrier-picker/SpreadBarrierCard.js
@@ -42,7 +42,7 @@ export default class SpreadBarrierCard extends Component {
                 <InputGroup
                     type="number"
                     label={`Amount per point (${currency})`}
-                    value={spreadInfo.amountPerPoint}
+                    defaultValue={spreadInfo.amountPerPoint}
                     onChange={amountPerPointChange}
                 />
                 <RadioGroup
@@ -54,13 +54,13 @@ export default class SpreadBarrierCard extends Component {
                 <InputGroup
                     type="number"
                     label="Stop loss"
-                    value={spreadInfo.stopLoss}           // TODO: hardcode default as backend return wrong data
+                    defaultValue={spreadInfo.stopLoss}           // TODO: hardcode default as backend return wrong data
                     onChange={stopLossChange}
                 />
                 <InputGroup
                     type="number"
                     label="Stop profit"
-                    value={spreadInfo.stopProfit}
+                    defaultValue={spreadInfo.stopProfit}
                     onChange={stopProfitChange}
                 />
             </div>
--- a/src/contract-details/ContractDetailsCard.js
+++ b/src/contract-details/ContractDetailsCard.js
@@ -10,10 +10,11 @@ export default class ContractDetailsCard extends Component {
 		nowEpoch: PropTypes.number,
 		transactionId: PropTypes.number,
 		actions: PropTypes.object,
+		durationUnit: PropTypes.string.isRequired,
 	};
 
 	render() {
-		const { contract, actions } = this.props;
+		const { contract, actions, durationUnit } = this.props;
 		const sold = !!contract.sell_price;
 		const validToSell = contract.is_valid_to_sell === 1 && !contract.is_expired;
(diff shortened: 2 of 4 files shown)


src/barrier-picker/SpreadBarrierCard.js
```diff
@@ -42,7 +42,7 @@ export default class SpreadBarrierCard extends Component {
                 <InputGroup
                     type="number"
                     label={`Amount per point (${currency})`}
-                    value={spreadInfo.amountPerPoint}
+                    defaultValue={spreadInfo.amountPerPoint}
```
qingweibinary: It uses value for a reason, to preserve state when change mode, we need to use value

src/contract-details/ContractDetailsCard.js
```diff
@@ -48,24 +49,25 @@ export default class ContractDetailsCard extends Component {
...
-						className="sell-at-market-btn"
-						onClick={() => actions.sellContract(contract.contract_id, 0)}
-					>
-						Sell at Market (<NumberPlain
-							value={contract.bid_price}
-							currency={contract.currency}
-							isProfit={v => v - contract.buy_price}
-						/>)
-					</button> :
+				{(durationUnit !== 't') && (
```
qingweibinary: Have you check that this actually happens?

By right, if it's not allowed to be sold, is_valid_to_sell would be 0, thus validToSell would be false, 

It should not required an additional field check for this.

qingweibinary:
I've checked, it's a backend issue, our current code is doing fine as long as we check on `is_valid_to_sell`
IMO, we should not decide the business rule in frontend unless necessary

Trello card of issue
https://trello.com/c/RZyeHDWh

src/barrier-picker/SpreadBarrierCard.js
nuruddeensalihu: Using value would fix the value though and therefore close it for modification or doesn't it ? It it need to use value why not use label ? Or am i confusing something ?

qingweibinary: Actions are used to change the value ~
We can bind actions on the input

src/contract-details/ContractDetailsCard.js
nuruddeensalihu: So if durationUnit is of tick value then , is_valid_to_sell would be 0 , is that what you are insinuating ? If that's the case i think its from the backend perhaps. The (durationUnit !== 't') is to check explicitly, with that i don't think sell on market would appear not even for a second .

src/barrier-picker/SpreadBarrierCard.js
nuruddeensalihu: Actions are used to change the value and from the same input group whose value has been fixed? Or do we have a different UI for that  ?

src/contract-details/ContractDetailsCard.js
qingweibinary: Yes ~
it's from backend and should be from backend, the durationUnit check would be some temporary workaround, as ContractDetail should not be deciding what to render by itself

src/barrier-picker/SpreadBarrierCard.js
qingweibinary: Have a look on other barrier?

Specifically the onStopLossChange and onStopProfitChange are actions that suppose to change the value of Input

nuruddeensalihu: I have seen all those. The value have to changed before they are fired right ? How can the value change when you fix the value by setting the value ?  Check again .

src/contract-details/ContractDetailsCard.js
borisyankov: After we start having these conditionals, even if they are correct, we need to extract to a 'biz logic' function/lib somewhere + unit test them

PR #188, opened 2016-03-28
Fix most pending test, except one that require more information for d…

qingweibinary:
…ebug


src/trades/__tests__/singleTradeSelectors-test.js
```diff
@@ -22,7 +23,7 @@ describe('singleTradeSelectors', () => {
         account: fromJS({ currency: 'JPY' }),
         ticks: fromJS({ R_9000: [{ quote: 123 }] }),
         trades: fromJS([{ symbol: 'R_9000' }]),
-        tradingOptions: fromJS([]),
+        tradingOptions: new Map({ R_9000: contractsForR_50 }),
```
borisyankov: fromJS({}) will create a map anyways.

qingweibinary: I know, but we have more problem on that XD
TradingOptions need some work, now it's a immutable object containing mutable js object, so using fromJS will convert everything to immutable and breaks.

I did not make big change to solve it altogether yet as some selectors depends on the mutable version

PR #190, opened 2016-03-28
Fix hide sell at market button when durationunit is tick

DanSallau:
Hi,

Attached contained few changes that should fix our or hide sell at market button when the duration unit is ticks. Should there be any issue with regard to this, i would be glad to address. Trello https://trello.com/c/0aPKfajT/31-sell-at-market-shouldn-t-be-shown-for-trades-with-duration-as-ticks

### DIFF
--- a/src/contract-details/ContractDetailsCard.js
+++ b/src/contract-details/ContractDetailsCard.js
@@ -10,10 +10,11 @@ export default class ContractDetailsCard extends Component {
 		nowEpoch: PropTypes.number,
 		transactionId: PropTypes.number,
 		actions: PropTypes.object,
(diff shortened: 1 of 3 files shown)


qingweibinary:
Just have a new finding, the problem I mentioned in https://trello.com/c/RZyeHDWh happens on more than 1 trading type

So the problem exists for not only tick but other trade, I think binary.com does not suffer from this because we do not show portfolio right after purchase contract, so the problematic api response is dropped.

This fix will only work for tick, maybe we should not fix it on our side

qingweibinary:
Update, junbon is fixing this, so maybe we do not need this PR

PR #191, opened 2016-03-29
Fix TradesTabs disabled purchase vutton issue

DanSallau:
HI,

Attached contained a few changes that should address our trade bug here https://trello.com/c/IEzL2OS5/27-click-on-new-trade-button-purchase-button-is-disabled-in-the-trade-window. Should there be any issue with regard to this, please do let me know would be happy to address.

### DIFF
--- a/src/trades/TradesTabs.js
+++ b/src/trades/TradesTabs.js
@@ -17,7 +17,6 @@ export default class TradesTabs extends Component {
 
     render() {
         const { actions, activeTradeIndex, assetsIsOpen, contracts, trades, ticksForAllSymbols, currency } = this.props;
-        const activeTrade = trades[activeTradeIndex];
         return (
             <div className="trades-tabs">
                 <TabList
@@ -36,12 +35,12 @@ export default class TradesTabs extends Component {
                 <div key={index}>
                     { (index === activeTradeIndex) &&    
                         <FullTradeCard
-                            index={activeTradeIndex}
+                            index={index}
                             actions={actions}
-                            trade={activeTrade}
-                            marketIsOpen={assetsIsOpen[activeTrade.symbol].isOpen}
-                            ticks={ticksForAllSymbols[activeTrade.symbol]}
-                            contract={contracts[activeTrade.symbol]}
+                            trade={trade}
+                            marketIsOpen={assetsIsOpen[trade.symbol].isOpen}
+                            ticks={ticksForAllSymbols[trade.symbol]}
+                            contract={contracts[trade.symbol]}
                             currency={currency}
                         />
                     }
--- a/src/trades/TradesTabs.js
+++ b/src/trades/TradesTabs.js
@@ -31,15 +31,22 @@ export default class TradesTabs extends Component {
                         />
                     )}
                 </TabList>
-                <FullTradeCard
-                    index={activeTradeIndex}
-                    actions={actions}
-                    trade={activeTrade}
-                    marketIsOpen={assetsIsOpen[activeTrade.symbol].isOpen}
-                    ticks={ticksForAllSymbols[activeTrade.symbol]}
-                    contract={contracts[activeTrade.symbol]}
-                    currency={currency}
-                />
+            {trades.map((trade, index) =>
+
+                <div key={index}>
+                    { (index === activeTradeIndex) &&    
+                        <FullTradeCard
+                            index={activeTradeIndex}
+                            actions={actions}
+                            trade={activeTrade}
+                            marketIsOpen={assetsIsOpen[activeTrade.symbol].isOpen}
+                            ticks={ticksForAllSymbols[activeTrade.symbol]}
+                            contract={contracts[activeTrade.symbol]}
+                            currency={currency}
+                        />
+                    }
+                </div>
+             )}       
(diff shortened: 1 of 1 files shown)


borisyankov:
Are you sure this fixes the issue?
To me it seems you are doing the same thing as before but in a more convoluted way.
If this actually fixes the issue, can you explain why (I am puzzled :) )

nuruddeensalihu:
This should be the best solution to the issue to my understanding. If you check the fulltradeparams you would notice that we structure the two the params to fire the onAssetChange event on every componetWillMount life cycle. Our tradegrid component obeys this structure and therefore no issue with the grid structure. We assumed our asset change on every new trade creation . But the tradeTabs on the other hand does not create new instance of trade on button click. Therefore no new contract is subscribed to and hence the issue with disabled purchase trade button. Its assumed that asset didn't change. The new modification would make the creation of new trade instead of just switching the tabIndexes and hence a full render is performed. This solves our issue at the moment.

borisyankov:
Oh, I get the issue and how your code fixes it.
This should not have been the solution, though - too convoluted. We can just call the onAssetChange function manually.

nuruddeensalihu:
onAssetChange  was called in componentWillMount already which the tradeTabs goes through once and the tradeGrid calls on every  new trade. From our design we intended for the two components to call the event on every new Trade. I think adding it manually somewhere would result in duplication and unnecessary. So i thought enforcing rendering on every new Trade like the case is in tradeGrids should be the ultimate solution for this .

borisyankov:
Adding it manually somewhere is definitely much more obvious and close to the correct solution.
Iterating over the whole list and rendering, solves this issue almost 'randomly' - always code with the principle of 'least surprise'.

@qingweibinary any opinion on this?

nuruddeensalihu:
I don't think the iteration size would be that much. The trades depends on the trades open by user. Usually 2 3 4 or even less as the case is on the tabGrid itself.

qingweibinary:
I agree this code can be improve on the map part, to maintain the previous way

Here's a few ways

1. let other component (In our case, there's New Trade Button, and Asset Pickers) call subscriptions themselves

2. change the criteria for FullTradeCard to make new subscription, ie. check for trade ID and absence of proposal

3. modify createTrade actions creator so it will do yet another api call (it already is calling tradingOptions and ticks stream) ~

I think 3 is easiest in our case, no duplicated code ~ and since we always want to subscribe , so it should be fine.

borisyankov:
#3 seems best to me too..

@nuruddeensalihu 
My concern is not with code performance for sure.
It is only with code correctness and obviousness of it.

qingweibinary:
@nuruddeensalihu 
If you are going with option 3, remember to remove the existing call from FullTradeCard   =)

nuruddeensalihu:
After going with 3, at the very end i realise it would the aim of our params. On new trade button click a new trade is added and then a subscription or a proposal is made to the trade/contract. This require few parameters such the symbol, barrier , duration etc. These parameters are not set on new trade button click considering the params component is not even rendered at that moment. So i don't think we can go with that option. Would look further and implement an alternative should there be any.

borisyankov:
@qingweibinary Can you help @nuruddeensalihu with the #3 solution, please ;)

qingweibinary:
@borisyankov roger, I'll have a deeper look

qingweibinary:
@nuruddeensalihu was correct on the problem, but I think we can do something on it.

I think the true problem lies on the misuse of onAssetChange() in componentWillMount (sorry, my fault).

My idea is that we create a method that will generate default trade params, (we already have them, just need to compose into one), using trade and tradingOptions (aka contracts) object.

so we replace the onAssetChange() in componentWillMount with this method, and initial subscription can make use of this method as well to get the necessary params.

I think this is slightly more work, but probably more correct as componentWillAmount only want to get a decent default params.

@nuruddeensalihu @borisyankov thoughts?

nuruddeensalihu:
The problem we are having is not with the onAssetChange() or the subscription. The problem is componentWillMount is called once on full render. At the moment we don't perform render on new Trade  for the tradeTabs view hence we don't go through the componentWillMount cycle. So going by your explanation above, we would still encounter the same problem. tradeTabs view does not perform a render and hence not going through mounting stage. Also its important to note that, subscription comes after parameters are set .

qingweibinary:
The problem is componentWillMount is called once on full render. 
    At the moment we don't perform render on new Trade for the tradeTabs view hence we 
    don't go through the componentWillMount cycle.

The actual problem is we should not rely onAssetChange to set default params, which is needed for subscription. The proper flow should be something as follow

1. Call create new trade actions
2. Get ticks and contract needed for the trade
3. create a trade object using default in reducer, update the symbol of trade object
4. Use trade object created + contract obtained to create a set of default params
5. Use default params to update the trade object
6. Call updateProposalSubscription

Note: 2 to 6 happens in createNewTrade actions, so you see there's no need of hooking into component lifecycle.

Did I miss out anything? it seems to me that we could remove componentWillMount totally if go with this path since setting defaults are delegated to actionCreator

nuruddeensalihu:
Its ok , i think i found a solution.

PR #192, opened 2016-03-30
Should only depends on valid_to_sell, backend should not return confu…

qingweibinary:
…sing payload


nuruddeensalihu:
Who's the headless guy above ? :wink:

qingweibinary:
A bot named Coverall?

PR #193, opened 2016-03-31
Add close button on tradeTabs

DanSallau:
Hi ,

Attached contained few changes to address the issue here https://trello.com/c/pU3PzMDA/37-cannot-close-multiple-trade-windows-for-different-assets as reported by the QA. Should there be any issue with regard to this, please do let me know and would be glad to address.

### DIFF
--- a/src/_common/Tab.js
+++ b/src/_common/Tab.js
@@ -9,15 +9,19 @@ export default class Tab extends Component {
         showText: PropTypes.bool,
         text: PropTypes.string,
         onMouseDown: PropTypes.func,
+        showCloseIcon: PropTypes.bool,
+        onClose: PropTypes.func,
+
     };
 
     static defaultProps = {
         showText: true,
         showIcon: true,
+        showCloseIcon: false,
     };
 
     render() {
-        const { imgSrc, selected, showIcon, showText, text, onMouseDown } = this.props;
+        const { imgSrc, selected, showIcon, showText, text, onMouseDown, onClose, showCloseIcon } = this.props;
 
         return (
             <div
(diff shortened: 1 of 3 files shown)


src/_common/Tab.js
```diff
@@ -28,6 +32,11 @@ export default class Tab extends Component {
             >
                 {showIcon && imgSrc && <img src={imgSrc} />}
                 {showText && text && <span>{text}</span>}
+                {showCloseIcon &&   
+                    (<button className="close-btn" onClick={onClose}>
```
borisyankov: You don't need to put () around button

src/_common/__tests__/Tab-test.js
```diff
@@ -19,6 +19,16 @@ describe('<Tab />', () => {
         expect(wrapper.text()).to.equal('');
     });
 
+    it('should render tab with close button', () => {
```
borisyankov: Tests. Good.
Be more descriptive. This is almost like documentation.
Good text here allows us to not write comments in the actual source code.
In that case I might call it 'renders a close button when showCloseIcon is true'.

src/trades/TradesTabs.js
```diff
@@ -28,6 +28,11 @@ export default class TradesTabs extends Component {
                         <Tab
                             key={index}
                             text={trade.symbolName}
+                            showCloseIcon
```
borisyankov: Rename to 'closable' everywhere :)

src/trades/TradesTabs.js
```diff
@@ -28,6 +28,11 @@ export default class TradesTabs extends Component {
...
+                            onClose={ev => {
+                                actions.removeTrade(index);
+                                ev.stopPropagation();
```
borisyankov: What happens if you don't call stopPropagation. Is it necessary?

nuruddeensalihu: Used to adding those in js buttons to address event bubbling issue [https://developer.mozilla.org/en/docs/Web/API/Event/stopPropagation] here. Didn't try without it in react though . :wink:

PR #196, opened 2016-04-01
Fix asset trade button click bug

DanSallau:
Hi,

Attached contained few changes that should fix the issue reported here https://trello.com/c/UQVzXh5Y/47-if-i-open-multiple-assets-for-trading-like-random-25-50-and-100-index-random-25-index-is-displayed-twice for your perusal. Should there be any issue with this, would be glad to address.

### DIFF
--- a/src/asset-picker/AssetPickerItem.js
+++ b/src/asset-picker/AssetPickerItem.js
@@ -34,9 +34,16 @@ export default class AssetPickerItem extends Component {
 		return (
 			<tr tabIndex={0}
 				className={classes}
-				onClick={() => onSelect(symbol)}
+				onClick={(ev) => {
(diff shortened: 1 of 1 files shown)


src/asset-picker/AssetPickerItem.js
```diff
@@ -34,9 +34,16 @@ export default class AssetPickerItem extends Component {
 		return (
 			<tr tabIndex={0}
 				className={classes}
-				onClick={() => onSelect(symbol)}
+				onClick={(ev) => {
+					onSelect(symbol);
+					ev.stopPropagation();
```
borisyankov: You still don't need the stopPropagation, I am sure?

PR #201, opened 2016-04-04
Temporarily handle error from contract for

qingweibinary:


src/trade-params/FullTradeParams.js
```diff
@@ -456,6 +456,10 @@ export default class FullTradeParams extends Component {
                     shown={!!trade.proposalError}
                     text={trade.proposalError ? trade.proposalError.message.split(')')[1] : ''}
                 />
+                <ErrorMsg
+                    shown={!!trade.contractForError}
```
borisyankov: You don't need to convert to boolean.

qingweibinary: Sometimes I feel it's more obvious that I want it to be boolean though ~~
Please forgive my paranoid XD

PR #203, opened 2016-04-04
Set default value for spread type

qingweibinary:


src/trade-params/FullTradeParams.js
```diff
@@ -399,7 +399,7 @@ export default class FullTradeParams extends Component {
                 }
                 {showDigitBarrier &&
                     <DigitBarrierCard
-                        barrier={trade.barrier}
+                        barrier={+(trade.barrier)}
```
borisyankov: No need for ()

src/payout-picker/StakeCard.js
```diff
@@ -22,12 +22,14 @@ export default class StakeCard extends Component {
 
     stepUp() {
         const { amount, onAmountChange } = this.props;
-        onAmountChange({ target: { value: amount + step } });
+        const newAmount = +(amount) + step;
```
borisyankov: No need for ()

src/payout-picker/StakeCard.js
```diff
@@ -22,12 +22,14 @@ export default class StakeCard extends Component {
...
+        onAmountChange({ target: { value: newAmount.toString() } });
     }
 
     stepDown() {
         const { amount, onAmountChange } = this.props;
-        onAmountChange({ target: { value: amount - step } });
+        const newAmount = +(amount) - step;
+        onAmountChange({ target: { value: newAmount } });
```
borisyankov: Why no toString( ) here?

qingweibinary: Because it's already guarantee to be string from the caller ~

PR #209, opened 2016-04-05
Qingwei/market picker improvement

qingweibinary:


src/_common/MarketPicker.js
```diff
@@ -13,6 +13,26 @@ export default class MarketPicker extends Component {
 
 	render() {
 		const { marketTree, onChange, allOptionShown, showMarkets, value } = this.props;
+		const markets = Object.keys(marketTree)
+			.filter(market => !(showMarkets && showMarkets.includes(marketTree[market].display_name)));
```
qingweibinary: I did not factor it out as i think it's pretty ad-hoc ~

PR #210, opened 2016-04-05
Nuru/beautify start later

DanSallau:
Hi,

Attached contained few changes that should beautify our toggle switch as described in the card here https://trello.com/c/PBLotA4R/55-beautify-start-later-toggle-following-abdo-s-suggestion .  Note: I have commented few css styling instead of deleting . I chose to comment those pending decision and would the needful later.


nuruddeensalihu:
What happened to travis ? 😉

public/styles.css
```diff
@@ -758,7 +758,10 @@ td.right, th.right {
 td.row-id {
     font-weight: bold;
 }
-
+.row {
```
borisyankov: We already have this style in binary-styles that we use.
Also, almost always use rems for this in this case about 0.5rems

public/styles.css
```diff
@@ -851,14 +854,59 @@ input[type=number]::-webkit-outer-spin-button {
...
 }
 
 .trade-chart {
     height: 250px;
     margin-bottom: 1rem;
 }
-
+.onoffswitch {
+    position: relative; width: 90px;
+    -webkit-user-select:none; -moz-user-select:none; -ms-user-select: none;
```
borisyankov: user-select is already set to none for all elements in binary-styles

public/styles.css
nuruddeensalihu: I learnt we already have one. I only extend and modify the top and bottom margins. However, we add a new css class for that if necessary ?

PR #212, opened 2016-04-06
Clear token after signout

DanSallau:
Hi,

I think we should clear token while signing out . At the moment we don't hence the signout does not work. Sign out and try opening authenticated pages and see it would work. 😉 Trello https://trello.com/c/81521QiF/73-logout-signout-does-not-really-signout-provided-there-is-a-new-another-window-opened


borisyankov:
Merging but don't bother with other token-related stuff ;)

PR #213, opened 2016-04-06
Make Barrier type more obvious

qingweibinary:


public/styles.css
```diff
@@ -977,3 +977,9 @@ input[type=number]::-webkit-outer-spin-button {
 #profit-loss-label {
     font-size: large;
 }
+
+#barrier-type-msg {
+    font-size: medium;
```
borisyankov: Try to use rems. They are relative to font size, thus a bit dynamic.
Since our font size is about 15px (may be 13, 14, it varies) the margin-top would be 1rem.

qingweibinary: OK ~

PR #215, opened 2016-04-06
Change input to text to prevent binding error

qingweibinary:


borisyankov:
I actually reworked the .css so now we get a serious conflict ;)

What binding error?
If it is not going to work with an input type=checkbox, it makes sense for it to be completely custom control than being an input=text

qingweibinary:
The one I solved with a directory called workaround ~

You're right, I'll rework it into a button ~

PR #216, opened 2016-04-06
Enhance/refactor errorMsg

DanSallau:
Hi,

On my attempt to refactor our errorMsg component and to avoid inline operations . I made few changes to the files below. Should there be any suggestion with regard to this please let me know would be glad to address.

### DIFF
--- a/src/_common/ErrorMsg.js
+++ b/src/_common/ErrorMsg.js
@@ -1,5 +1,6 @@
 import React, { PropTypes, Component } from 'react';
 import M from '../_common/M';
+import errorToString from '../_utils/errorToString';
 
 export default class ErrorMsg extends Component {
 
@@ -9,7 +10,6 @@ export default class ErrorMsg extends Component {
 	};
 	render() {
 		const { shown, text } = this.props;
-        const strText = text.split(')').length > 1 ? text.split(')')[1] : text;
-		return shown ? <p className="errorfield"><M m={strText} /></p> : <span />;
+		return shown ? <p className="errorfield"><M m={errorToString(text)} /></p> : null;
 	}
 }
--- a/src/_common/ErrorMsg.js
+++ b/src/_common/ErrorMsg.js
@@ -7,24 +7,9 @@ export default class ErrorMsg extends Component {
 		shown: PropTypes.bool.isRequired,
 		text: PropTypes.string.isRequired,
 	};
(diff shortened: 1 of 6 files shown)


borisyankov:
Using state for this is definitely not correct.

src/_common/ErrorMsg.js
```diff
@@ -7,9 +7,9 @@ export default class ErrorMsg extends Component {
 		shown: PropTypes.bool.isRequired,
 		text: PropTypes.string.isRequired,
 	};
-
 	render() {
 		const { shown, text } = this.props;
-		return shown ? <p className="errorfield"><M m={text} /></p> : <span />;
+        const strText = text.split(')').length > 1 ? text.split(')')[1] : text;
+		return shown ? <p className="errorfield"><M m={strText} /></p> : <span />;
```
borisyankov: if not shown return null, and not <span />

src/_common/ErrorMsg.js
```diff
@@ -7,9 +7,9 @@ export default class ErrorMsg extends Component {
 		shown: PropTypes.bool.isRequired,
 		text: PropTypes.string.isRequired,
 	};
-
 	render() {
 		const { shown, text } = this.props;
-		return shown ? <p className="errorfield"><M m={text} /></p> : <span />;
+        const strText = text.split(')').length > 1 ? text.split(')')[1] : text;
```
borisyankov: Extract into a function in the _utils folder. Like errorToString and put this there, plust a couple of unit tests :)

borisyankov:
Two comments, rest is fine now :)

src/_common/ErrorMsg.js
```diff
@@ -1,15 +1,15 @@
...
 
 	static propTypes = {
 		shown: PropTypes.bool.isRequired,
 		text: PropTypes.string.isRequired,
 	};
-
 	render() {
 		const { shown, text } = this.props;
-		return shown ? <p className="errorfield"><M m={text} /></p> : <span />;
+		return shown ? <p className="errorfield"><M m={errorToString(text)} /></p> : null;
```
borisyankov: We don't need to translate errors. 
They are already translated on the server ;)

nuruddeensalihu: Some are not server errors. Just hard coded

PR #221, opened 2016-04-08
Allow null country value

qingweibinary:


src/_common/LabeledText.js
```diff
@@ -10,8 +10,8 @@ export default class LabeledText extends Component {
     };
 
     render() {
-        const { id, label, value = 'N/A' } = this.props;
-
+        const { id, label } = this.props;
+        const value = this.props.value || 'N/A';
```
borisyankov: Noooooo ;)
Add 'N/A' in defaultProps => value

qingweibinary: does it work for null?

src/settings/SettingsDetails.js
```diff
@@ -21,7 +21,7 @@ export default class SettingsDetails extends Component {
 					<LabeledText
 						id="name"
 						label="Name"
-						value={settings.name}
+						value={settings.name || ''}
```
qingweibinary: @borisyankov this checking is required somewhere since it's null, I think placing them in SettingsDetails is better, as it's within context so we can have a little inconsistency.

What do you think?
it's return from backend as null, so I think we better keep it this way in store.

borisyankov: Should we also send null to the server then, when user updates and the value is ''?

qingweibinary: You're right, I'll store undefined instead ~

PR #224, opened 2016-04-08
Fix NOK fxTrade sell at and Goback bugs

DanSallau:
Hi,

Attached contained few changes that should address the issue reported here https://trello.com/c/44VJcERz/165-nok-forex-breaks-trade for your perusal. 

regard;

### DIFF
--- a/src/fulltrade/FullTradeCard.js
+++ b/src/fulltrade/FullTradeCard.js
@@ -67,7 +67,7 @@ export default class FullTradeCard extends Component {
                         <FullTradeParams
                             {...this.props}
                             disabled={disabled}
-                            contract={disabled ? mockedContract : contract}
+                            contract={contract}
                         />
                 }
             </div>
--- a/src/trade-params/TradeParamsValidation.js
+++ b/src/trade-params/TradeParamsValidation.js
@@ -1,10 +1,9 @@
-const epochWithinRange = (epoch, range) =>
-    !!(range.find(r => {
-        const millis = epoch * 1000;
-        const startMillis = r.open[0].getTime();
-        const endMillis = (r.close[1] || r.close[0]).getTime();
-        return millis >= startMillis && millis <= endMillis;
-    }));
+const epochWithinRange = (epoch, range) => {
+    const millis = epoch * 1000;
+    const startMillis = range.open[0].getTime();
+    const endMillis = (range.close[1] || range.close[0]).getTime();
+    return (millis >= startMillis && millis <= endMillis);
+};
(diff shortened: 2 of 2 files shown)


borisyankov:
@qingweibinary you are more familiar with this code. What is your opinion on this PR?

src/fulltrade/FullTradeCard.js
```diff
@@ -67,7 +67,7 @@ export default class FullTradeCard extends Component {
                         <FullTradeParams
                             {...this.props}
                             disabled={disabled}
-                            contract={disabled ? mockedContract : contract}
```
qingweibinary: Looks suspicious, why removing it?

src/trade-params/TradeParamsValidation.js
```diff
@@ -1,10 +1,9 @@
...
-        const startMillis = r.open[0].getTime();
-        const endMillis = (r.close[1] || r.close[0]).getTime();
-        return millis >= startMillis && millis <= endMillis;
-    }));
+const epochWithinRange = (epoch, range) => {
+    const millis = epoch * 1000;
+    const startMillis = range.open[0].getTime();
+    const endMillis = (range.close[1] || range.close[0]).getTime();
+    return (millis >= startMillis && millis <= endMillis);
+};
```
qingweibinary: Why change the interface?
Did it breaks in some other part of code?

src/fulltrade/FullTradeCard.js
nuruddeensalihu: Nothing suspicious there 😉 . I don't think it should be there to begin with. { disabled }  is partly determined by server response , hence mockedContract is often passed regardless of the value of contract . Moreover , we performed similar check on contract above .

src/trade-params/TradeParamsValidation.js
nuruddeensalihu: This breaks because in this case, range is not an array . `export const durationValid = (duration, durationUnit, options) =>
    !!options.find(opt => durationUnit === opt.unit && duration <= opt.max && duration >= opt.min);
`

src/fulltrade/FullTradeCard.js
qingweibinary: You're right ~

src/trade-params/TradeParamsValidation.js
qingweibinary: I see, I should've added test for this ~

@borisyankov I think it's ok to merge

PR #229, opened 2016-04-11
Retain duration parameters on type change

DanSallau:
Hi,

Attached contain few changes that should retain our duration parameters on type change as discussed in the comments here https://trello.com/c/O5Xcf0tB/56-digits-type-trade-is-not-working-and-broken-with-so-many-console-errors.

### DIFF
--- a/src/trade-params/FullTradeParams.js
+++ b/src/trade-params/FullTradeParams.js
@@ -218,7 +218,7 @@ export default class FullTradeParams extends Component {
 
         const category = trade.tradeCategory;
         const newDuration = createDefaultDuration(contract, category, newType);
-        const { dateStart, duration, durationUnit } = newDuration;
+        const { dateStart, duration, durationUnit } = trade || newDuration;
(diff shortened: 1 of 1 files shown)


borisyankov:
@qingweibinary opinions on the PR?

qingweibinary:
I think it is safe. The assumption here is that trade type under same category can share the same barrier, duration, which I think is true.

nuruddeensalihu:
@qingweibinary "it might breaks onAssetChange" . Do you mean like throw an error ? If so, i don't think it will. onAssetChange we create new trade therefore create new defaultBarrriers instead of the old trade . It does not break here as well.

borisyankov:
I am merging it, keep an eye on the potential issues :)

qingweibinary:
@nuruddeensalihu what I mean is here
https://github.com/binary-com/binary-next-gen/blob/master/src/trade-params/FullTradeParams.js#L147

It should not use old value from trade object when they are invalid, but this commit makes it to use values from trade object without checking. so we might provide invalid default params, it probably will not break the app in the way u describe, but it will break some functionalities we added.

PR #234, opened 2016-04-12
Fix trade duration issue

DanSallau:
Hi ,

Attached should fix a possible issue in our duration that may arises as noticed here https://trello.com/c/44VJcERz/165-nok-forex-breaks-trade .

### DIFF
--- a/src/trade-params/FullTradeParams.js
+++ b/src/trade-params/FullTradeParams.js
@@ -129,7 +129,6 @@ export default class FullTradeParams extends Component {
         const { contract, trade } = this.props;
 
         const selectedCat = trade.tradeCategory;
-
         if (!categoryValid(selectedCat, contract)) {
             this.onCategoryChange(Object.keys(contract)[0]);
         } else {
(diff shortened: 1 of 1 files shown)


borisyankov:
To me this code starts looking messier and messier.
We shouldn't rush bug fixes at the cost of clarity and good code.
Not sure how correct or optimal the approach is in this case... 
@qingweibinary - ok to merge or comments?

nuruddeensalihu:
This should come along with the fix yesterday . Didn't notice until Qinwei pointed out. It would break without this actually.

qingweibinary:
I think this is related to the pr I commented on #229

this PR solve it using some duplication, which I think is acceptable since we already abstract them.
To further abstract we might get some funny function like

`getDefaultBarrierAndDuration`

the code is very clean but I think less to do with this PR, more to the original structure.

qingweibinary:
One way to make it more concise (but maybe a bit arbritary)

is to give onTypeChange a parameter to decide whether to use old value or not

@nuruddeensalihu @borisyankov what do you think?

borisyankov:
Merging. Plan for bigger refactoring soon :)

PR #240, opened 2016-04-13
Qingwei/pass correct param to chart

qingweibinary:


src/fulltrade/FullTradeCard.js
```diff
@@ -42,6 +43,9 @@ export default class FullTradeCard extends Component {
             trade.disabled ||
             (!marketIsOpen && !contractAllowStartLater);
 
+        // TODO: remove usage of adapter so we have a consistent model
```
borisyankov: Yup.

PR #243, opened 2016-04-14
Nuru/fix asset selection

DanSallau:
Hi,

Attached contained some changes that would address some of our trade issues, for your perusal. The issues are as reported https://trello.com/c/WajP2doJ/189-getting-subscription-limit-message-too-often and https://trello.com/c/tPduKcu6/187-multiple-trade-windows-a-blank-window-is-shown-while-closing-them-one-by-one. There are quite few refactoring that should enhance and make our trade page better.


src/_actions/WorkspaceActions.js
```diff
@@ -42,6 +42,11 @@ export const changeActiveTrade = (activeTradeIndex) => ({
     activeTradeIndex,
 });
 
+export const setAssetBtnClicked = (assetBtnClicked) => ({
```
qingweibinary: parenthesis not required

src/trade-params/FullTradeParams.js
```diff
@@ -94,13 +94,19 @@ export default class FullTradeParams extends Component {
     componentWillMount() {
         this.onAssetChange();
     }
-
+    componentDidMount() {
+        const { actions, index, trade } = this.props;
+        if (!trade.hasOwnProperty('proposal')) {
```
borisyankov: Really no need to overdo the hasOwnProperty.
There is no prototype chain here.
if (!trade.proposal) is enough

nuruddeensalihu:
@borisyankov the reason behind the travis test failing would be address by qinwei PR. I noticed we now have a single page trade component(i.e trade TradeLayouts) and and made some changes to accommodates it as well. I think we still need this PR because it comes with some benefit. It saves us some unnecessary contract subscriptions, reduce the contract limit issue reported by bijan. and seems to speed up the trade a little bit in the process. Unless otherwise , i think its necessary.

qingweibinary:
@nuruddeensalihu try merge to clear both conflict and test failure

qingweibinary:
@nuruddeensalihu
this PR aim to solve repeated price proposal subscription, right ?

Does it still happen now, since we no longer have new trade button, I just test and find out we do not subscribe more than necessary when we change layout. But we have some other subscription issue though, I am checking.

The problem with this PR is the use of an seems to be random props, and it now couple the FullTradeCard component to some random component, for this case it's TradesLayout

It means now if someone want to use FullTradeCard, it needs to handle price proposal separately, which is pretty random ~

PR #244, opened 2016-04-14
Qingwei/asset picker

qingweibinary:


qingweibinary:
It looks awkward in a wide screen, maybe we can rearrange the assets so user do not need to scroll so much (if at all)

PR #247, opened 2016-04-15
Perform checking before render

qingweibinary:


qingweibinary:
Why clear localstorage fix the problem?
Because after clear, workspace will not try to render right panel, which will be broken if trying to render on load before data is available ~~

src/asset-details/AssetDetailsCard.js
```diff
@@ -33,7 +33,7 @@ export default class AssetDetailsCard extends Component {
 				}
 				<br />
 				<h5><M m="Durations" /></h5>
-				<AssetDetailsDurations durations={durations} />
+				{durations && <AssetDetailsDurations durations={durations} />}
```
borisyankov: Should the AssetDetailsDurations handle this?
In render if !durations return null

borisyankov: Hint: the question is rethorical :dancer:

qingweibinary: I think both would work, would you explain your reasoning and maybe we can generalize it as a rule so we have less decision to make =) ?

PR #248, opened 2016-04-18
Add null check on all TradeLayout

qingweibinary:


src/_data/LiveData.js
```diff
@@ -56,15 +56,14 @@ export const changeLanguage = langCode => {
...
 const initUnauthorized = async () => {
-
-};
-
-const initAuthorized = async (authData, store) => {
     api.getActiveSymbolsFull();
     api.getTradingTimes(new Date());
     api.getAssetIndex();
     api.getServerTime();
     api.getCandlesForLastNDays('R_100', 30);
```
qingweibinary: @borisyankov do you remember why do we move all api call after authorize?
Does it improve initial loading time?

borisyankov: It does, absolutely :)

borisyankov: In fact, next thing will be to load TradingTimes on first TradingTimesContainer render.
Same for everything that we do not depend on. Though I think we depend on trading times :( and asset index and may be few more calls.

qingweibinary: Yes, those dependencies cause the sign in to fail, because the redirect happens before data is ready.

since ws is async, does making these api before hand make a big difference?

I'll revert this part though

PR #250, opened 2016-04-18
Qingwei/trade lifecycle

qingweibinary:
@borisyankov 

This PR is not handling asset symbol, so it always create new trade of R_100, I think it's sensible because it's not clear what symbol should we use.

Is it fine?


borisyankov:
Yeah, R_100 is good for now. Later we might improve.

PR #251, opened 2016-04-18
Qingwei/show chart for statement

qingweibinary:
I move the ContractDetailsContainer (The Modal containing Chart) into workspace, and let statement and portfolio use the same instance. 

I named it singleton, which may or may not be appropriate, so please comment


src/_singletons/ContractDetailsSingleton.js
```diff
@@ -0,0 +1,29 @@
+import React, { PropTypes, Component } from 'react';
+import { connect } from 'react-redux';
+import Modal from '../containers/Modal';
+import ContractDetailsContainer from '../contract-details/ContractDetailsContainer';
+import ContractDetailsSingletonSelector from './ContractDetailsSingletonSelector';
+
+@connect(ContractDetailsSingletonSelector)
+export default class ContractDetailsSingleton extends Component {
```
borisyankov: Possibly Container is better name than Singleton.
Singleton is describing a bit of its purpose, but not rally the pattern, as I can certainly create more than one of it ;)

borisyankov:
Apart from the Singleton naming, I think its good ;)

src/_singletons/ContractDetailsSingleton.js
qingweibinary: OK, container already being used, I'll come up with another name.

And fix the test

qingweibinary:
@borisyankov updated and test passed

PR #252, opened 2016-04-18
Add trades on load if needed

qingweibinary:


src/trades/TradesLayouts.js
```diff
@@ -9,10 +9,16 @@ export default class TradesLayouts extends Component {
         currency: PropTypes.string.isRequired,
         ticksForAllSymbols: PropTypes.object.isRequired,
         trades: PropTypes.array.isRequired,
+        tradesCount: PropTypes.number.isRequired,
         layoutN: PropTypes.number.isRequired,
         contracts: PropTypes.object.isRequired,
     };
 
+    componentWillMount() {
```
borisyankov: Does not seem to be the correct place to do this.

borisyankov:
In this case, it seems to me that it will be a good idea to split into two actions.
CHANGE_ACTIVE_LAYOUT and CHANGE_TRADE_COUNT
CHANGE_ACTIVE_LAYOUT will call the CHANGE_TRADEC_COUNT, but on start, just CHANGE_TRADE_COUNT will be executed.

src/trades/TradesLayouts.js
qingweibinary: Why? since layout is only relevant to TradeLayout, I think associate them make sense. We only want to generate those additional trade object when we want to render TradeLayout, right?

If you're talking about componentWillMount, I had the same concern, but after some research, I find it's fine, the only downside is that it will render one more time

qingweibinary:
For your suggestion of splitting the action, I failed to see the point.
seems they always are related, right ?

PR #253, opened 2016-04-18
Fix deposit button bugs

DanSallau:
Hi,

Attached contained few changes that should address our Deposit or payment withdrawal bugs for your perusal. We are going to add an api call sendVerficationEmail() to our live api call .

### DIFF
--- a/src/deposit/WithdrawalForm.js
+++ b/src/deposit/WithdrawalForm.js
@@ -6,66 +6,73 @@ import InputGroup from '../_common/InputGroup';
 import SelectGroup from '../_common/SelectGroup';
 import Modal from '../containers/Modal';
 import currencies from '../_constants/currencies';
+import * as LiveData from '../_data/LiveData';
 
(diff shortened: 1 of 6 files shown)


src/deposit/WithdrawalForm.js
```diff
@@ -6,66 +6,73 @@ import InputGroup from '../_common/InputGroup';
...
-        confirmClicked: PropTypes.bool,
-        dryRunFailed: PropTypes.bool,
-        dryRunError: PropTypes.string,
-        inProgress: PropTypes.bool,
+        email: PropTypes.string,
     };
-
+    async componentWillMount() {
+        const { email } = this.props;
+        await LiveData.api.sendVerificationEmail(email);
```
borisyankov: componentWillMount() is not the place for these types of calls.

nuruddeensalihu: Thought as well. But considering it does not return a response. Where should we put it ?

PR #254, opened 2016-04-19
Qingwei/asset picker improve

qingweibinary:
The resubscription issue is surprisingly easy to solve ~


src/_common/Tab.js
```diff
@@ -21,13 +21,13 @@ export default class Tab extends Component {
...
     render() {
-        const { imgSrc, selected, showIcon, showText, text, onMouseDown, onClose, closable } = this.props;
+        const { imgSrc, selected, showIcon, showText, text, onClick, onClose, closable } = this.props;
 
         return (
             <div
                 role="tab"
                 aria-selected={selected}
-                onMouseDown={onMouseDown}
+                onClick={onClick}
```
qingweibinary: @borisyankov 

Do you remember why onMouseDown was use?
It causes a subtle bug on dropdown, explained as follow.

    When you click on children of dropdown, some event is fired on 
    the dom element you clicked, then if the dom element 
    shrink in sizes, suddenly the event will be fired on the 
    full-screen-overlay, thus trigger dropdown onclose method.

After an hour of searching the web, accidentally I found this code (god bless) 
do you think this changes will break anything?

nuruddeensalihu: @qingweibinary sounds like one of the events is propagating or isn't it ?  Perhaps you should stopPropagation and preventDefault on events . 😉

qingweibinary:
The trade type picker is still very broken in terms of UI, but I fixed the functionalities.

Here's the screenshots

### Basic
screen shot 2016-04-19 at 4 29 43 pm

### Digits
screen shot 2016-04-19 at 4 29 51 pm

### Advanced
screen shot 2016-04-19 at 4 29 59 pm


There's at least 2 issues
1. The size is way too different
2. The digits tabs should not show digits again, since that's the only category.

What's your take?

src/_common/Tab.js
qingweibinary: From my understanding, nothing to do with bubbling, which is already handled in DropDown component.

It's gotta do with how frequent the browser capture a click / mousedown event

When using mousedown, things changes once you press, and before you release (ie. mouseup) the dom changes, and thus detecting mousedown again in another element, then you release, which is a mouseup, so the 2nd mousedown combine with mouseup creates a click event, which mess things up 

I've tried abusing stop propagation and it did not work.

borisyankov: Click = mouse down + mouse up if there was very little mouse move between the two.
I'd say Click is generally better and more widely used, but mouse down fires earlier and results in faster experience for the user. Click is fired after the mouse up.

It is generally OK to replace Mouse Down with Click anywhere where it is causing an issue. It is also generally good to try to do the reverse for other cases to see if it will feel faster and make a difference.

borisyankov:
We will not show the subcategory anywhere.
Digits will be 3 rows 2 options now, show all options. As example it will be like

## Basic

| Rise | Fall |
| Higher | Lower |
| Stays In | Stays Out |
....

PR #255, opened 2016-04-19
Trade view redesign

DanSallau:
Hi,

Attached contained few suggestion on our trade view design for your perusal. I noticed the existing one is quite tight and small for the charts in some cases. Trello card here https://trello.com/c/eh5nGNiB/210-trade-view-redesign with snapshots .

### DIFF
--- a/public/styles.css
+++ b/public/styles.css
@@ -945,10 +945,6 @@ input[type=number]::-webkit-outer-spin-button {
 	
 	border-right:.5rem solid blue; 
 }
-
-.duration-card {
-    position: static;
-}
 .trade-chart-container {
     overflow: hidden;
     flex: 1;
--- a/public/styles.css
+++ b/public/styles.css
@@ -674,10 +674,6 @@ tr[tabindex] {
     visibility: hidden;;
 }
 
-.trade-panel {
-    flex: 1;
-}
-
 .toggle-buttons .btn-secondary.checked {
     background: rgba(255, 255, 255, .8);
 }
@@ -724,7 +720,7 @@ td.row-id {
     font-weight: bold;
 }
 .row-spacer {
-    margin: .5rem 0;
+    margin: 1rem 0;
 }
 
 .trade-header {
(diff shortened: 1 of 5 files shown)


public/styles.css
```diff
@@ -911,11 +910,41 @@ input[type=number]::-webkit-outer-spin-button {
...
+    margin: 1rem .5rem 1rem .5rem;
+    /*border: 1rem 1rem 1rem 1rem solid #73AD21 !important;*/
+    border: .01rem solid #eee !important;
+}
+.trade-panel {
+    -webkit-flex-direction: column;
+    flex-direction: column;
+    flex-flow: column
+}
+.arrow-down {
```
borisyankov: We already have these implemented in binary-styles or as svgs.
Doing the border hack as arrow is still a hack that does not allow a lot of things (like padding etc.)

src/trade-params/FullTradeParams.js
```diff
@@ -396,28 +396,46 @@ export default class FullTradeParams extends Component {
             <div className="trade-params" disabled={disabled}>
                 <DropDown
                     shown={trade.showAssetPicker}
-                    onClose={() => this.updateHelper('showAssetPicker', false, false)}
+                    onClose={(e) => {
+                            e.preventDefault();
```
borisyankov: We should rework this in the way I did it for Trade Type, not in that direction. Have a look.

src/trade-type-picker/TradeTypeDropDown.js
```diff
@@ -34,10 +33,14 @@ export default class TradeTypeDropDown extends Component {
                     />
                 </DropDown>
                 <div
-                    className="picker-label"
-                    onMouseDown={() => this.setState({ dropdownShown: true })}
+                    onMouseDown={(e) => {
+                        e.preventDefault();
```
borisyankov: Do we need this?

borisyankov:
Good initiatiive, but there are some plans on how to improve these components and half of the redesign does not follow them (because we haven't finished planning and I haven't explained them)

borisyankov:
If you had these as 5 smaller PRs I would have accepted maybe few, like this we need to discuss for longer time. Moving to SASS for the CSS is higher priority.

PR #257, opened 2016-04-20
Convert ContractDetails into List of key value pairs

qingweibinary:
WIP


src/contract-details/ContractDetailsCard.js
```diff
@@ -18,54 +19,28 @@ export default class ContractDetailsCard extends Component {
 		const workAroundForBackendBugIsDigitTrade = contract.shortcode && contract.shortcode.includes('DIGIT');
 		const validToSell = contract.is_valid_to_sell === 1 && !contract.is_expired;
 		const validationError = contract.validation_error;
-		const potentialProfit = contract.payout - contract.buy_price;
-		const profit = sold && (contract.sell_price - contract.buy_price);
+		const potentialProfit = (contract.payout - contract.buy_price).toFixed(2);
```
borisyankov: toMoney() may be or something? which will be equal toFixed(2) and later can be changed?
I suspect any of our internal amounts is toFixed(2)

qingweibinary: From my more than 10 years of experience in using money, I've yet encounter money using more than 2 decimal places 

=P

But for the sake of clarity, I'll add the method to utils and use it here

borisyankov: Haha :)
It is more to differentiate much more easily the usage of say 'you have $20.00' and 'this stock is $33.0023'.

Actually, this is probably incorrect assumption on your side. Japanese Yen maybe does not have significant numbers after the decimal point. Also, bitcoins are used with more symbols.

qingweibinary: Good point

PR #258, opened 2016-04-20
Debounce price proposal by 500 ms

qingweibinary:
WIP, 
1. we need to disable purchase when debouncing ~ 
2. Seems rate limit error is somewhat misleading, need a closer look


borisyankov:
The code is OK and ready to merge.
I am questioning the value of 500.
Did you test enough to see if debounce is what we need?
Keep in mind that 500 debounce means half a second delay of requesting a price, making the app work slower! This is important to not introduce just like that.

qingweibinary:
Yes, that's why it's WIP, i shove this PR up in the last 40 mins yesterday =P

I need to do some calculation

qingweibinary:
I think it's ok to merge, binary.com is using 500ms too, and so far no complain

borisyankov:
Since when binary-static is the gold standard? :)

qingweibinary:
Not sure where you read that =P

but my standard is `so far no complain`

borisyankov:
I mean if 250 is good enough we should use that. Or if 100 is good enough... got the idea? A smaller number is better for the user, larger for us.

qingweibinary:
OK, this requires some calculation and help from kaveh ~ 
I'll do it later

borisyankov:
👍 
'Help from Kaveh' can be in the form of raising limits too ;)

qingweibinary:
I talked about loosen up the limits with him 2 weeks ago, the problem is that someone might abuse it. 

I think technically it's possible to open up the limit for Next-gen, but we can discuss that part later ~

```
"max_proposal_subscription": {
        "max": 5,
        "applies_to": "subscribing to proposal concurrently"
      },
      "max_requests_pricing": {
        "minutely": 60,
        "hourly": 3600,
        "applies_to": "proposal and proposal_open_contract"
      },
      "max_requests_outcome": {
        "minutely": 60,
        "hourly": 3600,
        "applies_to": "portfolio, statement and proposal"
      },
      "max_requestes_general": {
        "minutely": 240,
        "hourly": 14400,
        "applies_to": "rest of calls"
      }
```

For price proposal, it's 60 call per minutes (1 per sec), of coz user not gonna trigger price proposal throughout the whole minute, I think we can make it 300 (tested, it's fine, but still breaks sometime, should be handle in another PR)

PR #259, opened 2016-04-20
Add support for sass to css compilation using grunt engine

DanSallau:
HI,

Attached contained few changes that should add sass to css compilation support for our project . I use grunt and is able to keep watching for a change on the sass files and compile them automatically when ever necessary. I plugin we intended to use previously did not work expected hence the switch to  grunt. Should this be satisfactory, would proceed and write .scss files.


borisyankov:
We are certainly not going to use Grunt to compile SASS. At worst it will be Gulp that we already use!

Another thing though is that we need to consider how will it be live-compiled. This needs to happen from the server.js file. The npm start command is doing a 'node server.js'.

qingweibinary:
By the way, have you tried this?

https://github.com/jtangelder/sass-loader

nuruddeensalihu:
@qingweibinary sass-loader does a similar thing as css-loader by webpack team. They both compile sass into css and inject them in to bundle.js. So basically you need an extractor to remove it from the bundle.js . Also for this to work , we need to reference our style style script into our js like `require('./style.sass)` and then the css-loader or sass-loader would translate them into style.css and inject it into bundle.js.  We can then extract from bundle.js. This i don't think is what we need.

PR #260, opened 2016-04-21
Translations

qingweibinary:
I'll try come out a way to automatically copy paste


qingweibinary:
It's ok to merge, have not found a good and easy way to do the automation

PR #261, opened 2016-04-21
Nuru/sass gulp

DanSallau:
Hi,

Attached contained few changes that would enable us to use gulp for our sass to css compiling and watching for your perusal. Should there be any issue with regard to this , please let me know 😉


borisyankov:
The approach should have been to integrate the gulp into our build (which currently it isn't) and use webpack for the live serving.

nuruddeensalihu:
"and use webpack for the live serving." Sorry it isn't clear. Can you elaborate more ? I don't seem to get it any more @borisyankov .

borisyankov:
I think we can do my suggestion later, in another PR.

PR #262, opened 2016-04-21
Qingwei/asset picker multi col

qingweibinary:


qingweibinary:
OK to merge

src/_common/KVColumn.js
```diff
@@ -0,0 +1,30 @@
+import React, { Component, PropTypes } from 'react';
+import M from './M';
+import Number from './Number';
+
+export default class KVColumn extends Component {
```
borisyankov: KV = KeyValue? No need to abbreviate.

PR #264, opened 2016-04-25
Qingwei/fix pickers size

qingweibinary:


qingweibinary:
This can be merged, trade type picker will be fix in another PR

PR #265, opened 2016-04-25
Fix trade Layout design

DanSallau:
Hi,

Attached contained few changes that should fix our trade layout design and as mentioned on this card. Also to make the layout consistent across other browsers.

https://trello.com/c/Zh0anvX2/222-fix-trade-layout-design


borisyankov:
'few changes that should fix our trade layout'... Maaan, these are 400 lines of new CSS, a lot of it repetitive. This is not few.

The good thing - a lot of layouts are better.
The bad thing - this is messy code, totally can not merge this!

public/styles.css
```diff
@@ -994,154 +994,496 @@ input[type=number]::-webkit-outer-spin-button {
...
 }
-
 .layout-mini {
     background-color: white;
     border-radius: 1px;
     margin: 1px;
 }
 
-.layout-1-1 {
+.layout-mini-1-1 {
```
borisyankov: That is not how we want to do it.
The mini button and the big layout work in the exact way, and we should be able to use the same styles. This change adds 300-400 lines of styles!!!

public/styles.css
```diff
@@ -994,154 +994,496 @@ input[type=number]::-webkit-outer-spin-button {
...
-.layout-5-5 > div:nth-child(4),
-.layout-5-5 > div:nth-child(5) {
+.layout-mini-5-5 > div:nth-child(4),
+.layout-mini-5-5 > div:nth-child(5) {
     flex-basis: 40%;
 }
 
+/*  Trade flex layout */ 
+.layout-1-1 {
+    -ms-flex-direction: row;
```
borisyankov: -ms and other prefixes will be added by a build step.

src/trades/TradesLayouts.js
```diff
@@ -26,22 +28,43 @@ export default class TradesLayouts extends Component {
 
     render() {
         const { actions, assetsIsOpen, currency, layoutN, trades, ticksForAllSymbols, contracts } = this.props;
+        const tradesCount = trades.length;
+        const contSize = tradesCount > 0 ? layoutData[tradesCount][layoutN] : undefined;
```
borisyankov: I am not sure what this code is, can you explain?
Also, can you think of another way to implement the styling without touching this file?

borisyankov:
@nuruddeensalihu This is a hard problem to fix, I know, if it wasn't I would've done it already.
Do not do such a big PR with so many changes, which I know you put a lot of work into, so lets make it work.
Lets work through each layout that needs fixing, one PR at a time.

Now, using this code, try a new PR, fixing first the layouts with minimal style changes and possibly no changes to React files.
If such changes are needed, we'll discuss and see if there is no other choice.

nuruddeensalihu:
@borisyankov what i wanted to do is a rewrite of the layout by simply separating the the trades in to two individual flex-containers in a big parent flex-container. that way we can achieve what all the layout we wanted in a more clear and flexible way. I understand that the mini layouts are to use the same style as the main layouts. However, i separate them here to have a more flexibility in setting our main layout basis. Also was trying to preserve the previous styling pending decision. 

The styling above does all we wanted in a flexible and responsive way. Read through once more and you would see. Yes the PR is long because we declare the respective container size in an array in a new js file tradeLayoutData . I was thinking if the new method is ok by you we can proceed and remove the repetitions where necessary as you said. I am still convince the way to achieve our layouts is by separating the trades into two/one children flex-containers and controlling their respective directions as it's done above.

PR #268, opened 2016-04-27
Qingwei/trivial fixed

qingweibinary:


src/asset-picker/AssetPickerFilter.js
```diff
@@ -19,10 +20,10 @@ export default class AssetPickerFilter extends Component {
 		actions.updateAssetPickerFilter(e);
 	}
 
-	// componentDidMount() {
-	// 	const assetSearchNode = findDOMNode(this.refs.assetSearch);
-    //     setTimeout(() => assetSearchNode.focus(), 1000);
-    // }
+    componentDidMount() {
+		const assetSearchNode = findDOMNode(this.refs.asset);
+        setTimeout(() => assetSearchNode.firstChild.focus(), 300);
```
qingweibinary: setTimeout is a hack ~ I was not able to find which part of the app got the focus, seems to be <body>, but I dunno why

Any idea?

public/styles.css
```diff
@@ -236,6 +236,10 @@ nav.sidebar {
 }
 
 .full-screen-overlay {
+    z-index: 9998;
```
borisyankov: Are you sure this is needed? It shouldn't be.
z-index is relevant only for relative and absolutely positioned things The content of full-screen-overlay are not using this, so they do not need to be z-indexed.

src/asset-picker/AssetPickerFilter.js
```diff
@@ -32,7 +33,7 @@ export default class AssetPickerFilter extends Component {
 		return (
 			<div className="asset-picker-filter">
 				<InputGroup
-					ref="assetSearch"
+					ref="asset"
```
borisyankov: Hmmmm, why the rename, seems a bit error prone?

src/asset-picker/AssetPickerFilter.js
borisyankov: Is it working with setTimeout though? It wasn't for me.

public/styles.css
qingweibinary: I see, I got it wrong, will fix

src/asset-picker/AssetPickerFilter.js
qingweibinary: OK, will name it back

src/asset-picker/AssetPickerFilter.js
qingweibinary: Because of firstChild ~

qingweibinary:
@borisyankov fixed , test passed

PR #274, opened 2016-05-02
Qingwei/dynamic data fetching

qingweibinary:
No intention to have it merged yet ~

I created a separated ChartData reducer keyed by contract_id

I am treating Chart in workspace and Chart in popup differently.
So BinaryChart that needs to show historical contract specific data will get from this reducer instead of the TicksReducer, what do you think?


borisyankov:
Not a fan in this approach. Even the commit from last time in ChartDataActions.js is not ideal.
We want all this logic to be isolated.

Think about it:
1. BinaryChart says 'I need OHLC for yesterday'
2. DataRetriever gets a message, returns data.
3. LiveApi is called (or not, maybe it is optional) by DataRetriever.

After we have this working perfectly for binary-chart, after that we will have:
* DataRetrieved is cached
* Optionally in Redux, and/or by LiveApi or DataRetriever.

If we don't do it like this, this functionality will not work for anything outside of next-gen and will need a ton of code added. Also, this will lead to looser coupling between components.

borisyankov:
Another thing related to ChartDataActions...
Determining what needs to be retrieved, and the actual api call is not necessarily the responsibility of the same component (though it might be, just a food for thought)

qingweibinary:
We have different opinion, but I am not sure where exactly since this span through a few project, so I'll try to make it clear

In my understanding

### Binary Chart
* Chart should not deal with data, the only thing it should do with data is to render chart using data it received.
    * This allow chart to be more reusable, albeit less usable.
* Since chart does not handle data by itself, it will takes in callback, and perform whatever it's caller want it to do.
* This means caller need to handle data themselves, so in next-gen, we will use reducer, in binarybot, amin will decide.


### Next-gen
* Next-gen should handles data to be pump into chart
   * Case 1: Workspace, data per symbol
   * Case 2: Statement/Portfolio, data per contract
* We can either (No strong opinion)
   1. generalize data fetching to work on symbol, 
   2. specialize so that it also works on contract-id


**So which part do we have conflict?** or anything that I left out.
One part I think is 

    If we don't do it like this, this functionality will not work for anything 
    outside of next-gen and will need a ton of code added.

Which come to the question that how much do we want to support, I think we need to pick 2 from 3 below
1. More code
2. Flexible
3. Easy to use

and letting user to implement their own data retrieval logic seems fine to me, as it is tightly couple to data storing, which chart should not dictate over.

borisyankov:
We agree on:
* Binary Charts should not handle data directly.
* It should just 'ask' via callback for more data, and get it asynchronously.

You were wondering about:
* We definitely do not want to handle data differently for the ticks and contracts.
* We want 2 and 3

Not exactly agreeing:
* Next-gen should not need to know too much about that

Meainging, the DataRetriever (which might use a better name ;) ) will not be in Binary Charts, it might be in Live-Api, it might not be. It will be the reusable thing that by default can use binary-live-api to ask for more data.

Caching can be done via Redux or not.

I think a consumer of BinaryCharts should be able to plug his own instance of binary-live-api and ge tthe data (without caching).

BinaryCharts should work without live data too, ticks=... and ohlc=... (just as it works with ticks now).

qingweibinary:
Right now `DataRetriever already is in `LiveApi`, and not `Binary Chart`

So the things I need to change will be 

* to unify ticks and contract data handling in next-gen layer

right?

Storing data is responsibility of next-gen instead of Live Api, right ?

qingweibinary:
This PR depends on https://github.com/binary-com/binary-live-api/pull/40

I suggest do not merge until it's tested with updated dependencies ~
I think we cannot fully unify Chart on `FullTradeCard` vs `ContractDetailContainer`
although both are ticks data (maybe candle later), but they are keyed differently (symbol vs contract_id)

If you have good way to unify them, let me know, and I'll change accordingly

qingweibinary:
@borisyankov I think this is ok to merge ~

PR #276, opened 2016-05-03
Qingwei/fix contract details

qingweibinary:


qingweibinary:
This should be OK to merge, **trade type picker** is more tricky, I'll do it in another PR

PR #277, opened 2016-05-04
Qingwei/refactor trade params

qingweibinary:
WIP, do not merge =)


qingweibinary:
I think this is ok to merge ~ 

I did some test with different layout and changing assets, duration etc

PR #278, opened 2016-05-04
Fix firefox layout issue

DanSallau:
Hi,

Attached contained few changes that should address our FF layout issue. Would test more and look for possibility of improvement and should there be any ,  would be glad to improve on that.

### DIFF
--- a/src/layouts/layouts.css
+++ b/src/layouts/layouts.css
@@ -70,6 +70,7 @@
 
 .layout-2-1 > div {
     flex: 1;
+    max-height: calc(96% / 2);
 }
 
 /* 2-2 */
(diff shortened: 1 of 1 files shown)


src/layouts/layouts.css
```diff
@@ -90,6 +91,7 @@
 
 .layout-3-1 > div {
     flex: 1;
+    max-height: calc(96% / 2);
```
borisyankov: Why is this not just 48% ?

nuruddeensalihu: Oh , i used the calc for both of them bcos i thought it would be easier to divide an odd number 😉

PR #281, opened 2016-05-04
Qingwei/refactor trade params

qingweibinary:


qingweibinary:
This make Trade Type Picker in a separate view in mobile ~~

Not awesome, but works ;)

PR #282, opened 2016-05-05
Clear asset picker search query onClose

DanSallau:
Hi,

Attached contained some changes for resetting our search queries as explained here https://trello.com/c/dUCC72AX/245-asset-picker.

### DIFF
--- a/public/styles.css
+++ b/public/styles.css
@@ -926,6 +926,9 @@ input[type=number]::-webkit-outer-spin-button {
     position: absolute;
     z-index: 10000;
     box-shadow: 0 0 2rem 0 black;
+   
+    /* Re position dropdown to avoid layout break*/
+    top: 7%;
 }
 
 .picker-label {
--- a/src/asset-picker/AssetPickerDropDown.js
+++ b/src/asset-picker/AssetPickerDropDown.js
@@ -35,26 +35,27 @@ export default class AssetPickerDropDown extends Component {
             router.push('asset-picker');
         }
     }
+    onClose(e) {
+        const { actions } = this.props;
+        e.stopPropagation();
+        actions.updateAssetPickerSearchQuery('');
+        actions.updateAssetPickerFilter('all');
+        this.setState({ dropdownShown: false });
+    }
 
     render() {
         const { actions, index, selectedSymbol, selectedSymbolName } = this.props;
         const { dropdownShown } = this.state;
         return (
             <div>
-                <DropDown
-                    shown={dropdownShown}
-                    onClose={() => this.setState({ dropdownShown: false })}
-                >
+                <DropDown shown={dropdownShown} onClose={::this.onClose}>
                     <AssetPickerContainer
                         actions={actions}
                         tradeIdx={index}
                         selectedAsset={selectedSymbol}
                     />
                 </DropDown>
-                <div
-                    className="picker-label"
-                    onMouseDown={::this.openAssetPicker}
-                >
+                <div className="picker-label" onMouseDown={::this.openAssetPicker}>
                     {selectedSymbolName}
                 </div>
             </div>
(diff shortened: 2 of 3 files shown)


public/styles.css
```diff
@@ -926,6 +926,9 @@ input[type=number]::-webkit-outer-spin-button {
     position: absolute;
     z-index: 10000;
     box-shadow: 0 0 2rem 0 black;
+   
+    /* Re position dropdown to avoid layout break*/
+    top: 7%;
```
borisyankov: Is there any reasoning about the 7%, NEVER use random numbers. Unlikely to work everytime.

src/asset-picker/AssetPickerDropDown.js
```diff
@@ -35,26 +35,27 @@ export default class AssetPickerDropDown extends Component {
             router.push('asset-picker');
         }
     }
+    onClose(e) {
+        const { actions } = this.props;
+        e.stopPropagation();
```
borisyankov: Absolutely no need to stop propagation.

src/asset-picker/AssetPickerDropDown.js
```diff
@@ -35,26 +35,27 @@ export default class AssetPickerDropDown extends Component {
...
+        actions.updateAssetPickerSearchQuery('');
+        actions.updateAssetPickerFilter('all');
```
borisyankov: Add a new action instead called resetAssetPickerFilter.
We should keep double state access for exceptional cases, and not as a rule.

src/payout-picker/StakeCard.js
```diff
@@ -56,7 +56,6 @@ export default class StakeCard extends Component {
                     min={0.35}
                     max={100000}
                     step="0.01"
-                    list="amounts"
```
borisyankov: Why remove? I already fixed that.

src/asset-picker/AssetPickerDropDown.js
nuruddeensalihu: if we add a new rule just to do the same thing. Wouldn't it be like doing the same thing with just a different name? I mean we have a rule that can just do that for us. The query is set my the same rule, modified and can as well clear by the same rule can't it?.

src/asset-picker/AssetPickerDropDown.js
nuruddeensalihu: Our drop is on a a modal with some higher z-index and drown above other controls. Closing it involved clicking any part of the page away from the dropdown . Wouldn't b a good practice to stop propagation in case of any which in this case is possible? :wink:

public/styles.css
nuruddeensalihu: I think with the said percentage its safer instead of using a defined number. I am not defining a height for the dropdown and i don't think there should be an instance where the said 7% would not be enough. When you defined a top, it always measure from the parent container viewport. However, in this case i only measure and assumed. Should there be more accurate number we can use perhaps ? 😉

src/payout-picker/StakeCard.js
nuruddeensalihu: Oh didn't know you fix that . Its showing dropdown in a supposed textbox control which are not necessary. Nonetheless, I think the list params shouldn't be provided there or should it?

src/asset-picker/AssetPickerDropDown.js
borisyankov: No, because first, it will explain the action much better.
Also, two actions cause two evaluations of all reducers, which is possibly a perf concern.

public/styles.css
borisyankov: bottom: 0 seems like a better solution to me.
Or bottom: .5rem even better.

src/asset-picker/AssetPickerDropDown.js
borisyankov: In that case, this can be added to the overlay in DropDown component, but not here.

borisyankov:
If this PR is not needed, close it.

PR #284, opened 2016-05-09
Reset Search asset picker query

DanSallau:
Hi,

Attached contained few changes that should address our asset search reset query issue. Should there be any concern with regard to this, would be glad to address.


borisyankov:
Looks good.

PR #285, opened 2016-05-09
Fix drop-down height

DanSallau:
Hi,

Attached contained few styling changes that should address our drop down height issue like below.

Before 

image

With the new PR 

image

Should there be any issue with the PR above, would be glad to address.


borisyankov:
This was not a great fix. Always try to fix with more general solution. 
Also, moving the picker up is not ideal.
Fixed in a commit in different way.

PR #286, opened 2016-05-09
Qingwei/refactor trade object

qingweibinary:


qingweibinary:
@borisyankov do not merge until you read my slack pm ~
We need some other solution

qingweibinary:
The last commit will fix the race condition mentioned https://trello.com/c/LgTpaPbx
The problem happens due to many small actions being triggered once Component is mounted, and we unmount the component before all actions are completed.

qingweibinary:
@borisyankov this PR is ready, it breaks the chart after you purchase, but this is more correct, the updateExtremes call should not depends on trade and contract changes, as conceptually, Trade and Contract object does not change, at least from POV of chart, they should not change.

Meaning updateExtremes should be called on other places.

Problem ~

screen shot 2016-05-09 at 12 40 08 pm

The extremes is set wrongly (I believe when contract are 1st added into the chart) then it is never updated again since contract and trade values remain the same.

Let me know if you think it's this PR's mistake instead of binary-charts

qingweibinary:
I forgot to test one part, I'll close it to prevent accidental merge

qingweibinary:
Ok to be merge now

src/_reducers/TradesReducer.js
```diff
@@ -9,18 +9,27 @@ import {
...
+        symbol: 'R_100',
+        tradeCategory: 'risefall',
+        duration: 5,
+        durationUnit: 't',
+        basis: 'stake',
+        amount: 50,
+        type: 'CALL',
+        barrierType: 'relative',
+    },
+    uiState: {
```
borisyankov: this does not sound like good name.

src/_reducers/TradesReducer.js
```diff
@@ -9,18 +9,27 @@ import {
...
+        showAssetPicker: false,
```
borisyankov: This is now a local state for the picker and nothing to do with Reducers I think.

src/_reducers/TradesReducer.js
qingweibinary: suggestion? things inside are like `showAssetPicker` which is totally ui state

src/fulltrade/FullTradeCard.js
```diff
@@ -33,32 +36,34 @@ export default class FullTradeCard extends Component {
...
+        const { symbolName } = trade.params;
 
         const contract = this.props.contract || mockedContract;
 
         const contractAllowStartLater = findDeep(contract, child => child && !!child.forwardStartingDuration);
 
         const disabled =
             contract === mockedContract ||
-            trade.disabled ||
+            trade.uiState.disabled ||
```
borisyankov: I think you are moving in the wrong direction.
We need to make a better differentiation than a 'trade panel' and a trade object.
Trade object will be flat and has only stuff you need to perform the trade.
Trade panel can be what you are trying to do here.

src/_reducers/TradesReducer.js
qingweibinary: Oh, did not notice, I'll verify and remove as needed

qingweibinary:
@borisyankov 

I removed showAssetPicker state, I suggest we merge it 1st, and open another card for structural changes if needed, I think it would means splitting TradesReducer into 2 to 3 reducers, and wire them up in selectors, which should not affect most changes here, as the actions' interface will be the same.

PR #289, opened 2016-05-10
Qingwei/improve perf

qingweibinary:
This is still not exactly correct, reason being we have nested structure, and converting from immutableJS to plainJS breaks equality.

Good news is that, although the program is not exactly correct, the re-render are still **close to correct**, as the whole state tree TradesContainer is listening to does not change unless there's new ticks.

However, when there's multiple trade, Cannikin Law applies, the most frequently updated trade will make other trade updates too ~ But it does not trigger reflow as React will do it's own diff


qingweibinary:
I will try to run a VM to simulate slow machines, and see how it goes

qingweibinary:
In ubuntu box, 1 CPU, 768mb RAM, the page feel laggy ~
switching between layout takes up to 4 seconds 
and multiple trade panel is close to unusable 

will upgrade the vm and test later

PR #294, opened 2016-05-11
Redesign toogle button

DanSallau:
Hi ,

Attached contained  few changes that should fix our toggle button . 

image

image

with PR 

image

image

Card https://trello.com/c/l3P5VfAC/261-bug-start-later-toggle-button-is-messed-up.

Would proceed and take out the text content from the css later this week.


nuruddeensalihu:
@borisyankov have realised the styling has been modified and the changes were introduced to the toggle button style where not needed. Not sure why thought, but above is it new and correct look. Let me know if we are going to keep or discard it ?

PR #296, opened 2016-05-11
Qingwei/optimize

qingweibinary:
Do not merge ~~

@borisyankov
The open contract problem could not be solve unless we want to implement our own selectors.
And I am not convince all these changes actually improve things, do you know anyway we could verify ?

I did some profiling but could not really understand the data, from heap size, our master perform better as this branch did a lot of memoization, but I dunno how to analyse the CPU profile.


qingweibinary:
The 3rd commit ebdb984 manage to solve the re-render caused by open contract

But I havent test thoroughly yet ~ 
better not merge it and I'll do some profiling tmr

borisyankov:
Yep, it isn't well tested because there is no single unit-test ;)

borisyankov:
I will try with refactoring too, this is getting too hairy IMHO.

qingweibinary:
Added test ~ did a CPU profile using NODE_ENV=production

Does not show significant improvement, but does not re-render with open contract streaming.

borisyankov:
That is quite significant.
100ms on a Macbook Pro can easily be 200-500ms on a worse machine. Considering this happens for each trade, and each 1-2secs... definitely noticeable.

PR #297, opened 2016-05-11
Unsubscribe when changing symbols

qingweibinary:


nuruddeensalihu:
I think you need to close and reopen this PR so as to enable it start the travis build. 😉

qingweibinary:
@nuruddeensalihu Thanks for notifying =)

src/_reducers/TickReducer.js
```diff
@@ -9,6 +9,16 @@ import {
 
 const initialState = new Map();
 
+const smartMergeTicks = (existingTicks, newTicks) => {
```
borisyankov: 'smart' really? :)
Is that the most descriptive name :)

src/_reducers/TickReducer.js
```diff
@@ -18,22 +28,24 @@ export default (state = initialState, action) => {
                 epoch: +tick.epoch,
                 quote: +tick.quote,
             };
-            return state.update(symbol, List.of(), v => v.takeLast(60).push(newTick));
+            return state.update(symbol, List.of(), v => v.takeLast(4999).push(newTick));
```
borisyankov: Do we really want to get the maximum amount of data each time?
I am not sure. This is probably hundreds of kbs of data.

qingweibinary: this is the storing part, if we do not store all of them, the dynamic data loading would not work properly, 4999 certainly is an arbritary number where we could adjust, but it will still be large.

let's use this number and test it out ? better than wild guessing what is the best 

Or even better if you have alternative approach?

qingweibinary:
Reopen to trigger travis

PR #298, opened 2016-05-11
Replace modal with js alert

DanSallau:
Hi,

Attached contained few changes that should replace the modal error reported here with showError . https://trello.com/c/Q8HpcRGv/239-if-dark-theme-is-selected-then-alert-messages-pop-up-window-text-color-is-very-light.

I noticed the trade_panel color itself need changing when dark boris because its not clear enought. However, i am not sure what color to use or whats the plan for those. IF there is any you can let me know i would execute. Also the error message reported is the only message i found liable for this change without repercussion as most alter state or execute html.

### DIFF
--- a/src/fulltrade/FullTradeCard.js
+++ b/src/fulltrade/FullTradeCard.js
@@ -3,7 +3,6 @@ import shouldPureComponentUpdate from 'react-pure-render/function';
 import { BinaryChart } from 'binary-charts';
 // const BinaryChart = (props) => <div {...props} style={{ background: 'grey' }} />;
 import PurchaseFailed from '../_common/PurchaseFailed';
-import Modal from '../containers/Modal';
 import FullTradeParams from '../trade-params/FullTradeParams';
 import ContractReceiptCard from './ContractReceiptCard';
 import findDeep from 'binary-utils/lib/findDeep';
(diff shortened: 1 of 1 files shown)


src/fulltrade/FullTradeCard.js
```diff
@@ -90,12 +89,10 @@ export default class FullTradeCard extends Component {
...
             <div disabled={disabled} className={'trade-panel'}>
-                <Modal
-                    shown={!!purchaseInfo.buy_error}
-                    onClose={() => actions.updatePurchaseInfo(index, 'buy_error', undefined)}
-                >
-                    <PurchaseFailed failure={purchaseInfo.buy_error} />
-                </Modal>
+
+                {!!purchaseInfo.buy_error &&
+                    showError(purchaseInfo.buy_error.code + ' ' + purchaseInfo.buy_error.message)
```
borisyankov: Is this even working?
A render method needs to be pure.

nuruddeensalihu: Oh i forgot showError is not a component . Nonetheless @borisyankov wouldn't The current Modal be more appropriate for our need for this case? Have analyse through all the Modals and thought this qualifies for the showError util , considering its being render in the render method wouldn't using the modal be more appropriate , less we are going to take it to another method. Have thought a lot but the requirement not well clear.

borisyankov:
Yeah, possibly, maybe not the Modal, but some inline error message is better than showError.

PR #299, opened 2016-05-11
Split duration card and forward starting

qingweibinary:
I do not understand why duration and forward starting could block changes in layout, hope this PR would solve them problem.

They still share a few props.


nuruddeensalihu:
When you split does it still overlapped or has it taken care of the issue here https://trello.com/c/ixGNLZyl/268-trade-panel-view-is-overlapped-when-having-startlater ?

qingweibinary:
@nuruddeensalihu I do not think this would solve the problem, it just change structure, but not style

PR #300, opened 2016-05-12
Fix safari videos horizontal issue.

DanSallau:
Hi,

Attached contained few changes that should address the safari vid list issue noted here https://trello.com/c/Fq9zVAmL/246-cross-browser-testing

### DIFF
--- a/public/styles.css
+++ b/public/styles.css
@@ -571,11 +571,16 @@ select {
 
 .video-list {
     margin: 0;
+    flex-wrap: wrap;
+    display: flex;
+    flex-direction: row;
 }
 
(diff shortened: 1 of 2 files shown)


public/styles.css
```diff
@@ -571,11 +571,16 @@ select {
...
+    flex-direction: row;
 }
 
 .video-thumbnail {
-    margin: .5rem;
-    min-width: 10rem;
+    margin: auto;
+    min-width: 10em;
+    max-width: 23em;
+    float: left;
```
borisyankov: Float is certainly not needed

PR #301, opened 2016-05-12
Barriertype always absolute if it comes from contract

qingweibinary:


qingweibinary:
Unfortunately, barrier value from `proposal_open_contract` call is always `absolute`, as opposed to the `price_proposal` call.

Right now, I will just add `{ barrierType: 'absolute' } ` to contract before passing into charts.
We should think about discarding trade object or draw a clearer line between it's purpose compared to contract.

eg. Trade object is for visualize proposal to be bought, contract is for contract bought

If so we might need more data from price_proposal instead of just relying on trade object

borisyankov:
No worries. Cool thing is everything is so well documented on the back-end, nothing can go wrong :)

PR #304, opened 2016-05-16
Qingwei/improve contract retrieval

qingweibinary:


qingweibinary:
This pr depends on https://github.com/binary-com/binary-live-api/pull/45

Do not merge until live-api pr published, else the chart range selector will break

PR #305, opened 2016-05-17
Qingwei/ohlc

qingweibinary:
WIP, do not merge ~

It depends on binary-chart and live-api that's not published yet

### DIFF
--- a/src/_reducers/ChartDataReducer.js
+++ b/src/_reducers/ChartDataReducer.js
@@ -7,25 +7,26 @@ const initialState = fromJS({});
 export default (state = initialState, action) => {
     switch (action.type) {
         case UPDATE_CHART_DATA_BY_CONTRACT: {
-            const { contractID, data } = action;
-            const existing = state.get(contractID) ? state.get(contractID) : [];
+            const { contractID, data, dataType } = action;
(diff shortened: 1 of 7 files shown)


qingweibinary:
@borisyankov there's 3 cases where we use chart in next-gen

a) Render closed contract, render when user click on transactions in statement
b) Render open contract, render when user click on entries in portfolio
c) Render symbol , the most obvious chart

How should OHLC support act in these cases?
For a), its simple, since both ticks and candles data are static.

For b) and c), now we do not have the ability to stream OHLC, so I presume we will just get candles, but the chart will not stream, until user click back ticks

Let me know if you disagree

borisyankov:
These are not the only options.
A closed contract is rendered when you are looking at an open contract, either in full-trade or in the portfolio, and the contract expires or is sold. ;)

Up till now the logic in the other apps (also after discussion with Rakshit) was something like 'if the duration is more than 1-2mins) - show a candlestick. I think this is way too simplistic, and works much better being able to switch between options easily.

A huge error in current charting in static and webtrader is showing too may candlesticks. If you show more than 20-30 items (or other number, that we can experiment, may be a candlestick per 25 pixels) the benefit is lost and the chart looks like a franken-tick chart, since you almost can't make out the details ;)

PR #315, opened 2016-05-25
Qingwei/fix issue

qingweibinary:


src/trade-params/TradeParams.js
```diff
@@ -85,6 +85,10 @@ export default class TradeParams extends Component {
         this.onStopTypeChange = ::this.onStopTypeChange;
         this.onStopProfitChange = ::this.onStopProfitChange;
         this.onPurchase = ::this.onPurchase;
+
+        this.state = {
+            dynamicKey: 0,
+        };
```
qingweibinary: @borisyankov alert , I am remounting the whole dom whenever asset change, this is to reset default value of inputs, and the reason of using default value is to allow user to input continuously.

I understand this is bad, so if you know any better way, please let me know, React is particularly bad in this aspect >.<

public/styles.css
```diff
@@ -3,6 +3,10 @@ body {
     overflow: hidden;
 }
 
+button {
+    margin-right: .2rem;
```
borisyankov: Waaaat? All buttons? :)

qingweibinary: it's very small ~ XD

OK, I'll add class or use inline-style

src/contract-details/SellAtMarketButton.js
```diff
@@ -20,7 +20,7 @@ export default class SellAtMarketButton extends Component {
 				className="sell-at-market-btn"
 				onClick={onClick}
 			>
-				<M m="Sell at Market" />
+				<M m="Sell at Market " />
```
borisyankov: That will messs up translations.

qingweibinary: Exactly, I'll update po files

borisyankov: Why do we need the space though?

qingweibinary: Because there will be price next to this span, and it does not looks nice without spacing

src/trade-params/TradeParams.js
borisyankov: Yeeeah this is bad ;)
We can have the inputs controlled.
By setting defaultValue and just listening on changes we have them uncontrolled.
Setting value should work.

qingweibinary: Controlled input does not work unless we introduce another middle state, because we want to allow user to input let say `12`, but by having it controlled, user can only input `1` or `2` but never `12`.

I think Introducing another middle state is worse than this approach.

borisyankov: I am not sure why you think that. Elaborate?

src/contract-details/SellAtMarketButton.js
borisyankov: &nbsp; or <span>space here</span> look more correct approaches ;)

src/trade-params/TradeParams.js
qingweibinary: you mean another middle state? or you dont think controlled component will disallow user input

I'll reply both

### Why controlled Component does not work?
I am not 100% sure of the reason, I think it happens like this

1. User key in `1`
2. Component fire action
3. Component updated by receiving updated props
4. User key in `2`
5. go back to 2.

I think it has to do with chrome treating field in datetime as a whole, the whole field is always selected, as opposed to text field where you have a moving cursor.

### Why need another middle state ?
to handle the problem, we need to have a state which does not update store whenever user key in, but update the controlled component state, and then we might need to set some timer so that the user input finally propagate to store.

This sounds complicated enough in words.

Note: I just tried in firefox and the start later is not working at all, now I start to appreciate the dropdown in binary.com XD, we might need to use text input or some ui library >.<

src/contract-details/SellAtMarketButton.js
qingweibinary: Absolutely, amended

src/trade-params/TradeParams.js
borisyankov: Alright merging it for now.
Don't over test. I also think we can have something better than time and date picker.
Even something like a smart input like 'in 20 mins' or it might be too smart in a bad way :)
Just will need some time to think it through without the pressure.

PR #318, opened 2016-05-26
Qingwei/fix issue

qingweibinary:


qingweibinary:
WIP, do not merge

public/styles.css
```diff
@@ -1073,6 +1073,7 @@ input[type=number]::-webkit-outer-spin-button {
 .mobile-content .trade-params > div {
     margin-right: .5rem;
     min-width: 10rem;
```
nuruddeensalihu: I think we shouldn't set the min-width here , It stretch the control where not required and tend to compensate with spaces where the control cannot fit the req. I mentioned it my new PR . I am of the opinion that in this case , we set for a control individually where necessary.

qingweibinary: My reasoning is simple and possibly flawed, having min-width will make sure all control have similar sizes, which would looks better.

Is this reasoning sound? or is the current approach problematic?

nuruddeensalihu: I was thinking about the controls variations and forgot that the screen/trade size varies also. I think you have a point considering container width changes with the layout. Controls tend to stretch sometimes when minimum width is set on the parent container. This usually is the case with radio and checkboxes who tend to compensate the minimum width with margins in most cases. I think we are fine in this case now that i check 😉

public/styles.css
```diff
@@ -1073,6 +1073,7 @@ input[type=number]::-webkit-outer-spin-button {
...
+    max-width: 15rem;
```
borisyankov: Unlikely to be a good idea, unlikely to be a terrible idea ;)

PR #319, opened 2016-05-26
Fix Lowethan display

DanSallau:
Hi,

Attached contained few changes that should address some little design issue on our params as discuss on the card here https://trello.com/c/OyP3K3Fh/365-width-positions-of-some-params-control .


public/styles.css
```diff
@@ -1157,6 +1158,8 @@ input[type=number]::-webkit-outer-spin-button {
 
 .barrier-input {
     display: flex;
+    flex-direction: column;
+    max-width: 11.3em;
```
borisyankov: That looks awfully random.

PR #320, opened 2016-05-26
Qingwei/fix issue

qingweibinary:


qingweibinary:
Something is wrong with the git history, if it matters let me know, i'll do something with it

PR #328, opened 2016-05-30
Qingwei/unify error msg

qingweibinary:
@borisyankov 
Questions:
1. Do we want to differentiate between warning and error? 
   So far the only things that should be warning is when barrier value exceeding pipSize, which backend will do truncation.


qingweibinary:
I think my guess was right, now each component does not own their own error, it's more complex ....

PR #332, opened 2016-06-01
Nuru/codepush

DanSallau:
Hi,

Attached contained changes for the code-push feature . Have tested the cordova and the code-push itself. The new www is a replica of public folder and the later should be removed if possible. This PR contained the necessary configuration we need for the codepush and cordova . I have added a two commands in READ.me namely ' Setting up the App' and 'running the app'.


borisyankov:
Looks OK, but we will not want the www folder to duplicate files.
Two options:
1. Find a way to configure it.
2. If not, we will have the copy as a build step in Gulp.

nuruddeensalihu:
Considering its just the naming , www instead of public. Can't we just rename the public folder to www?

PR #337, opened 2016-06-01
Nuru/code push

DanSallau:
Hi,

Attached contained changes for the code-push feature . Have tested the cordova and the code-push itself.I have added the cordova flatform because i noticed in the mobile project they added as well. Think they do that to reserved version perhaps. For the plugins. Below is the list of plugins 

image

The easiest way is to create new cordova project, 

`cordova create tetsProj`

and then copy the plugins folder to your next-gen folder. then install 
`
cordova plugin add cordova-plugin-code-push@latest`

which is the only external plugin we manually installed so far. 

Check for cordova-plugin-whitelist plugin in your plugin list. If its not there, install and then we are good to go. 😉


www/index.html
```diff
@@ -18,17 +19,17 @@
             <img class="spinner" src="img/binary-symbol-logo.svg" />
         </div>
     </div>
-
-<script src="boot.js"></script>
-<script src="https://cdn.polyfill.io/v2/polyfill.min.js?features=fetch,Promise,Array.from,Array.prototype.find,Array.prototype.includes,String.prototype.startsWith,String.prototype.includes,Math.sign,Intl.~locale.en,Intl.~locale.id,Intl.~locale.ru,Intl.~locale.fr&unknown=polyfill"></script>
-<script async src="app.js"></script>
-
+<script type="text/javascript" src="cordova.js"></script>
```
borisyankov: Does this work for web?
We should not load it always.

nuruddeensalihu: @borisyankov we can separate the two then . I added a home.html for the cordova .

config.xml
```diff
@@ -130,11 +79,15 @@
     <allow-intent href="geo:*" />
     <platform name="android">
         <allow-intent href="market:*" />
+        <preference name="CodePushDeploymentKey" value="rVLju2NoxbcCyC9f3h2biGcCRx8Y4JBH-fLXW" />
     </platform>
```
qingweibinary: Are these key sensitive data?

qingweibinary:
Looks good to me

There's merge conflict @nuruddeensalihu

www/js/index.js
```diff
@@ -0,0 +1,51 @@
...
+    // deviceready Event Handler
+    //
+    // The scope of 'this' is the event. In order to call the 'receivedEvent'
+    // function, we must explicitly call 'app.receivedEvent(...);'
+    onDeviceReady: function() {
+        app.receivedEvent('deviceready');
+    },
+    // Update DOM on a Received Event
+    receivedEvent: function(id) {
+        var parentElement = document.getElementById(id);
```
qingweibinary: How does it works? seems to received `deviceready` as id, so it will try find a dom element with `deviceready` as id?

config.xml
nuruddeensalihu: I think they should not be. You can have the key but what matters is the login and password info without with you cant deploy to codepush. You can see on similar project here https://github.com/rangle/code-push-demo/blob/master/config.xml , and binary mobile here https://github.com/binary-com/mobile/blob/master/config.xml and many more. To my understanding, i don't what someone can do with the keys , you need authentication to login to codepush before you can deploy and with the authenticatio, you can retrieve these keys themselves. So they are not sensitive.

www/js/index.js
nuruddeensalihu: @qingweibinary this is a mere template and possibly where we are going to do some works after configuration and installation. Codepush has few api a]that enable us set rules for the app like 

--When (and how often) to check for an update? (e.g. app start, in response to clicking a button in a settings page, periodically at some fixed interval)

--When an update is available, how to present it to the end user?

These are documented here https://microsoft.github.io/code-push/docs/cordova.html.  The index.js can be our entry file or we can edit or change. There are many events also that enable to listen to the app statuses like when the app is ready(ondeviceryReady above) etc. 

This is just a mere template with example specifying the basic events we might need along the way. We would rework this to suit our need later on.

config.xml
qingweibinary: OK, that's fine

www/js/index.js
qingweibinary: OK, so it does not matter now then. Thanks

PR #343, opened 2016-06-07
Fix mobile asset reset issue

DanSallau:
Hi,

Attached contained few changes that should address the issue mentioned in the card here https://trello.com/c/3pYyA6yj/450-asset-picker-does-not-reset-filter-in-mobile for your perusal.

### DIFF
--- a/src/mobile/MobileToolbarBack.js
+++ b/src/mobile/MobileToolbarBack.js
@@ -6,21 +6,24 @@ export default class MobileToolbarBack extends Component {
 
 	static propTypes = {
 		backBtnBarTitle: PropTypes.string,
+		actions: PropTypes.object,
+		to: PropTypes.string,
 	};
-
+	goBack(e) {
+		const { actions } = this.props;
+		e.preventDefault();
+		actions.resetAssetPickerFilter();
+		history.back();
+	}
 	render() {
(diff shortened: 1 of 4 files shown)


src/mobile/MobileToolbarBack.js
```diff
@@ -6,21 +6,24 @@ export default class MobileToolbarBack extends Component {
...
 	static propTypes = {
 		backBtnBarTitle: PropTypes.string,
+		actions: PropTypes.object,
+		to: PropTypes.string,
 	};
-
+	goBack(e) {
+		const { actions } = this.props;
+		e.preventDefault();
+		actions.resetAssetPickerFilter();
```
borisyankov: That is really not correct.
You should not have to change this file at all.
resetAssetPickerFilter does not belong here.

nuruddeensalihu: resetAssetPickerFilter is existing for desktop version already. We use that to reset or refresh our asset filter or am i missing something ? CAn't we just call it and achieve the same thing for the mobile version ?

qingweibinary: @nuruddeensalihu I guess boris meant it should not be called here `Inside goBack`, but somewhere else, maybe in componentWillMount of relevant Component

PR #348, opened 2016-06-09
Qingwei/mobile style fix

qingweibinary:
Drastic changes, tested in all existing layout and mobile


qingweibinary:
WIP ~

src/duration-picker/DurationCard.js
```diff
@@ -101,9 +101,9 @@ export default class DurationCard extends Component {
         if (!currentUnitBlock) return null;
 
         return (
-            <div className="duration-picker">
+            <div className="param-row">
```
borisyankov: Nope, this is a step backwards.
duration-picker, something-else-picker {
...
}
is much better than param-row {
}
Since then you can also style separately.

qingweibinary: OK, how about making `param-row` css a mixin , and mix into all pickers?
Is this fine?

borisyankov: Yes, but use mixins very sparingly. In that case likely not needed.
As an example what a mess sass might become is girddle that we use as grid in binary-static

borisyankov:
Merge conflict.

qingweibinary:
Dont merge!

qingweibinary:
OK to merge

borisyankov:
Did latest commits make it unmergable? Plz fix again :)

qingweibinary:
Yes, fixing

PR #353, opened 2016-06-10
Working copy, asset and tradetype on chart

qingweibinary:


qingweibinary:
Do not merge

qingweibinary:
Do you think this is an improvement?

screen shot 2016-06-10 at 11 31 31 am

borisyankov:
Not really. I thought about it, but not really.
For example: 
Why exactly over chart?
Why under the range controls, logically they should be separate.
And also: the flow is for the user to select these two first, then trade params.
So from left to right and from top to bottom, moving.

PR #355, opened 2016-06-10
Fix statement between two virtual and real accounts

DanSallau:
Hi,

Attached contained few changes that should fix our statements sharing issue as mentioned on the card here https://trello.com/c/XmnLmHrF/400-test-real-accounts-new-accounts-switching-accounts for your perusal.


qingweibinary:
@nuruddeensalihu merge conflict

borisyankov:
There are two better approaches:
1. Super simple. Do not cache Statement (I already did that)
2. More involved, but better. Do separate transactions by account ID, so no need to delete them.

PR #357, opened 2016-06-10
Fix start later mobile height

DanSallau:
Hi,

Attached contained few changes that should address the mobile issue mentioned on the ticket here https://trello.com/c/Nk8Qbc6G/469-navigator-is-covered-by-trade-params for your perusal.

before PR

image

with the PR

image

Notice the chart navigation disappear in the before. This is because its suppress by the tradeparams.

### DIFF
--- a/www/scss/_mobile.scss
+++ b/www/scss/_mobile.scss
@@ -23,6 +23,30 @@
   .setting-container > .asset-index-card {
     margin: 0 .25em 0 .25em;
   }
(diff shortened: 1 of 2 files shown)


borisyankov:
Does that look better to you?
We can't just randomly make all text size 65%.
Have you tried it on a phone? This would be unreadable. 
Your Macbook's emulation is twice as large as actual phone.

PR #359, opened 2016-06-13
Fix standalone news display

DanSallau:
HI,

Attached contained few changes that should fix the news standalone display as mentioned on the card here https://trello.com/c/yhq0y0PG/499-fix-standalone-view-artcle for your perusal .

image


qingweibinary:
@nuruddeensalihu branch out of data

src/news/NewsCard.js
```diff
@@ -11,20 +20,51 @@ export default class NewsCard extends Component {
         router: React.PropTypes.object.isRequired,
     };
 
+    componentWillMount() {
+         this.setState({
+            showArticle: false,
+            params: { index: 0 },
+        });
```
qingweibinary: You do not need to initialize state in both constructor and componentWillMount

src/news/ArticlePreviewMobile.js
```diff
@@ -0,0 +1,21 @@
+import React, { PropTypes, Component } from 'react';
```
qingweibinary: ArticlePreview and ArticlePreviewMobile looks close enough, why have both?

nuruddeensalihu: Yup the only different between the two is that preview expect an onClick object. I have separated two actually because i thought bringing about a total separation of the two would be better considering the parents components have been separated .

qingweibinary:
@nuruddeensalihu branch went out of date ~

nuruddeensalihu:
@qingweibinary i think he can just click the update branch button when ready 😉 Continuos updating the branch at the moment would just keep adding unnecessary commits .

qingweibinary:
@nuruddeensalihu ok, an alternative is rebase ~ but use it with care =)

PR #365, opened 2016-06-17
Fix stake or amount in trades

DanSallau:
Hi,

Attached contained few changes that should address the following ticket for your perusal. https://trello.com/c/opiISpUY/533-make-default-params-of-each-trade-different

### DIFF
--- a/src/_reducers/trades/TradesParamsReducer.js
+++ b/src/_reducers/trades/TradesParamsReducer.js
@@ -37,7 +37,12 @@ export default (state = initialState, action) => {
 
             if (countDiff > 0) {
                 const additionalTradeParams = Repeat(fromJS(defaultParams), countDiff);   // eslint-disable-line new-cap
-                return state.concat(additionalTradeParams);
+                const updatedParams = additionalTradeParams.map((param, index) => {
+                    const newParam = param.set('amount', param.get('amount') - ((state.size + index) * 10));
+
+                    return newParam;
+                });
+                return state.concat(updatedParams);
             }
 
             if (countDiff < 0) {


borisyankov:
I don't think we want to do it this way.
Also, if the user changes the stake, he will still encounter the error.

qingweibinary:
Anyway, we should solve it now so it does not block us, I prefer to use different symbol/type rather than different stake

borisyankov:
Differing asset might make more sense, but also might not.
This should not be an error. Backend is wrong and can fix it very easily.
@qingweibinary Can you push around a bit, it is like a critical bug for us now.

qingweibinary:
@borisyankov ok, I am not 100% convince on how it should works, how about I fire an email discussion ?

nuruddeensalihu:
"if the user changes the stake, he will still encounter the error." I don't think user would like his second trade params to match the first param or would he ? If he shouldn't then the error message would serve it purpose , thought not sure if the error message match the error type. Moreover, I don't think different asset would make more sense here considering user need to key in an amount regardless. That should be more convenient since its a text control compare to the said asset which involved a long list on a dropdown.

qingweibinary:
@nuruddeensalihu we would not know, as he could actually just sum up stake in 1 trade.
Different asset make more sense because simply changing stake is the least unlikely what the client want

If I want a stake 30 and a stake 40, I would use stake 70, not having both 30 and 40.

I suggest we change asset now so that it's not blocking us, then we will work on the frontend caching as discussed in email I sent

borisyankov:
Nope. None of the options make sense.
A user often trades on a single asset at a time.
He also often trades in similar stakes, especially for ticks at minimal value of 0.35 so just giving him 10, 20, 30, 40 stake is not helping anything.

And of course the user can likely have the same trade, there is another variable that he can vary - time. He can buy the first now, and the second few moments later.

PR #369, opened 2016-06-21
Qingwei/fix rc

qingweibinary:
Do not merge yet, I am testing =)

### DIFF
--- a/package.json
+++ b/package.json
@@ -78,6 +78,7 @@
   "babel": {
     "presets": [
       "react",
+      "es2015",
       "es2015-webpack",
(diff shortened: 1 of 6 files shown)


qingweibinary:
This would fix test temporary, to use webpack with mocha is kinda indirect, you need to bundle the code then only you can run the test.
And it requires an additional file to state all the files for webpack entry

I think the old method is better

qingweibinary:
OK to merge

borisyankov:
Exactly the reason I did not add it earlier :)

PR #370, opened 2016-06-21
Close popup when pressed esc

qingweibinary:


borisyankov:
Sure you can extract this behavior in the DropDown component itself? 🎯

qingweibinary:
Yes I could

PR #373, opened 2016-06-27
Use latest stable version for source map to work

qingweibinary:


qingweibinary:
Guess this would not work, do you know what is our dependency on the beta webpack?

PR #375, opened 2016-06-27
Qingwei/tester issue

qingweibinary:
I feel it's not reliable, as the same code was not working for half an hour, any idea ?


borisyankov:
You don't seem to be handling the fact that the user can quit by clicking Esc or other ways?

qingweibinary:
@borisyankov
Argh, will fix

qingweibinary:
@borisyankov Added handling code ~

borisyankov:
It is better as a separate component. I'll do it.

PR #382, opened 2016-06-29
Nuru/change splash

DanSallau:
Hi,

Attached contained few changes that should replace our splash screen for your perusal. It should further clean our config as well.


qingweibinary:
@nuruddeensalihu test failed, just in case u did not notice

PR #383, opened 2016-06-30
Qingwei/refactor trade hierarchy

qingweibinary:
Dont merge


qingweibinary:
I am thinking of deferring immutableToJS call until close to leaf of component hierarchy, any comment ?

borisyankov:
Don't go overboard with refactoring ;)

src/_store/directSelectors.js
```diff
@@ -20,6 +21,12 @@ export const ohlcSelector = state => state.ohlc;
 export const tradingOptionsSelector = state => state.tradingOptions;
 export const statesSelector = state => state.states;
 
+export const tradeParamsSelector = state => state.tradesParams;
+export const tradeProposalInfoSelector = state => state.tradesProposalInfo;
```
borisyankov: Is porposal a better name than proposalInfo?

src/trade-params/TradeParams.js
```diff
@@ -44,6 +46,18 @@ import { changeAsset, changeBarrier1, changeBarrier2, changeCategory, changeStar
...
+const errorToShow = errorObj => {
+    const { barrierError, contractError, durationError, proposalError, purchaseError } = errorObj;
+
+    if (contractError) return contractError;
+    if (barrierError) return barrierError;
+    if (durationError) return durationError;
+    if (proposalError) return proposalError;
+    if (purchaseError) return purchaseError;
+
+    return undefined;
```
borisyankov: That is simpler expressed via something like:
return errorObj.find(x => x);

borisyankov: Also, do we care wha type of error is it?
Wouldn't a list of errors be enough?
Maybe not, just a question...

src/trade-params/TradeParams.js
```diff
@@ -56,12 +70,12 @@ export default class TradeParams extends Component {
         contract: PropTypes.object,
         compact: PropTypes.bool,
         disabled: PropTypes.bool,
-        error: PropTypes.string,
+        errors: PropTypes.object,
```
borisyankov: Try to not have objects as params, they will diff on shouldComponentUpdate.

src/trade-params/TradeParams.js
```diff
@@ -101,8 +115,8 @@ export default class TradeParams extends Component {
     }
 
     componentWillUnmount() {
-        const { proposal, proposalError } = this.props;
-        if (proposalError) {
+        const { proposal, errors } = this.props;
+        if (errors.proposalError) {
```
borisyankov: You probably can abstract away these in a lib...
Like canUpdateProposal() or something.
Didn't think it through much, but basically the biz logic if possible to go outside of the component itself.

src/trade-params/TradeParams.js
```diff
@@ -182,6 +196,11 @@ export default class TradeParams extends Component {
         this.updateTradeParams(updatedAmountPerPoint);
     }
 
+    onCloseModal = () => {
+        const { actions, index } = this.props;
```
borisyankov: Intending to move away from modals as possible.
Just FYI.
And possibly all errors will be shown in the form.
We can also try showing all errors, not just the first one.

src/trade-params/TradeParams.js
```diff
@@ -256,9 +275,14 @@ export default class TradeParams extends Component {
         const showSpreadBarrier = categoryToUse === 'spreads';
         const askPrice = askPriceFromProposal(proposal);
 
+        const errorText = errorToShow(errors);
+
         return (
             <div className="trade-params" disabled={disabled} key={this.state.dynamicKey} style={style}>
-                <ErrorMsg text={error} />
+                <Modal shown={!!errors.purchaseError} onClose={this.onCloseModal}>
```
borisyankov: Not sure if we want more Modals.

src/trade/TradeCard.js
```diff
@@ -169,7 +151,9 @@ export default class TradeCard extends Component {
             serverContractModelToChartContractModel(lastBoughtContract);
 
         // contract error is not tied to trade, but symbol, thus not in tradeErrors
-        const tradeError = (propsContract ? propsContract.error : undefined) || errorToShow(tradeErrors);
+        const tradeError = propsContract ?
+            Object.assign(tradeErrors, { contractError: propsContract.error }) :
```
borisyankov: The more I look at the code, the more I am in favor of having arrays, not objects.

src/trade/test/SingleTradeXCard.js
```diff
@@ -0,0 +1,18 @@
+import React, { Component } from 'react';
+import { connect } from 'react-redux';
+import shouldPureComponentUpdate from 'react-pure-render/function';
+import TradeXCardContainer from './TradeXCardContainer';
+import immutableChildrenToJS from 'binary-utils/lib/immutableChildrenToJS';
+import { singleTest } from './TradeParamsSelector';
+
+@connect(singleTest)
+export default class SingleTradeXCard extends Component {
```
borisyankov: What does TradeX mean?

src/trade/test/TradeParamsSelector.js
```diff
@@ -0,0 +1,89 @@
...
+    [assetsIsOpenSelector, paramPerTrade(index)],
+    (assetsIsOpen, param) => {
+        const symbol = param.get('symbol');
+        return assetsIsOpen[symbol].isOpen;
+    }
+);
+
+const errorPerTrade = index => state => tradesErrorSelector(state).get(index);
+const proposalPerTrade = index => state => tradeProposalInfoSelector(state).get(index);
+const uiStatePerTrade = index => state => tradesUIStatesSelector(state).get(index);
```
borisyankov: uiState is reeeeeeally not descriptive.

src/trade/test/TradeParamsSelector.js
```diff
@@ -0,0 +1,89 @@
...
+const purchasePerTrade = index => state => tradePurchaseInfoSelector(state).get(index);
+
+const getStartLaterOnlyContract = contract => {
+    const startLaterCategories =
+        filterObjectBy(contract, child =>
```
borisyankov: This function is one of the reasons I don't like tree-like structures for our core data formats. Something flatter like a table would likely work better.

borisyankov:
Keep an eye on very heavy selectors (too much computation being done).
I am not sure if this is an issue we have or not, just a warning.

src/_store/directSelectors.js
qingweibinary: Let me think about it, originally it is an object containing proposal and a few other stuff, maybe not the case now

src/trade-params/TradeParams.js
qingweibinary: Exactly, which is why I propose to defer immutableToJS, to retain the equality, sometime it's just makes more sense >.<

src/trade-params/TradeParams.js
qingweibinary: We do because we want to show more important error message

src/trade-params/TradeParams.js
qingweibinary: right now I am reluctant to change TradeParams since it's working fine, I was just trying to prevent unnecessary render

I'll go through it after refactor is mergeable

src/trade-params/TradeParams.js
qingweibinary: I think it's a good idea, just we need to do it in a way that UI does not breakdown in mobile

src/trade/TradeCard.js
qingweibinary: In this case, I think you are right, since we do not yet really care about the type of error. Will change

src/trade/test/SingleTradeXCard.js
qingweibinary: Experimental XD

src/trade/test/TradeParamsSelector.js
qingweibinary: Hmm, I cannot visualize it directly, let me think a bit, but yea, I think we under-utilize the Record type provided by immutableJS (I know it exists from your commit XD)

qingweibinary:
Computations have to be done somewhere, and it's better in selector than inside render, since reselect gives us a chance to trade memory with computations, I'll keep an eye, have been thinking

qingweibinary:
This is working, but dont merge 1st, I want to test on app then clean up old code

PR #384, opened 2016-07-03
Fix redirect, set orintation temporarily, edit content security policy

DanSallau:
Hi,

Attached contained few changes to address our redirect issue for your perusal . Please this is ongoing and would welcomed any comment/review. AT the moment, a user does not have to reauthorize after closing the app and opening again. However, it does not refresh/reload after we close the child webview.

### DIFF
--- a/config.xml
+++ b/config.xml
@@ -11,6 +11,8 @@
     <plugin name="cordova-plugin-whitelist" spec="1" />
     <plugin name="cordova-plugin-code-push" version="1.8.2-beta" />
     <plugin name="cordova-plugin-splashscreen" version="3.2.2" />
+    <plugin name="cordova-plugin-keyboard" version="1.1.4" />
+    <plugin name="cordova-plugin-inappbrowser" value="CDVInAppBrowser" />
     <allow-navigation href="*://*.binary.com/*" />
     <access origin="*" />
     <allow-intent href="http://*/*" />
@@ -20,6 +22,8 @@
     <allow-intent href="mailto:*" />
     <allow-intent href="geo:*" />
     <preference name="BackgroundColor" value="#2a3052" />
+    <preference name="KeyboardShrinksView" value="true" /> 
+    <preference name="orientation" value="portrait" />
     <preference name="SplashScreen" value="screen" />
     <preference name="SplashScreenBackgroundColor" value="#2a3052" />
     <preference name="AutoHideSplashScreen" value="true" />
--- a/src/_data/Auth.js
+++ b/src/_data/Auth.js
@@ -30,13 +30,47 @@ export const signout = (nextState, replace) => {
     store.dispatch(updateToken(''));
     replace({ pathname: '/signin', state: nextState });
 };
+const parseOAuthResponse = (responseUrl) => {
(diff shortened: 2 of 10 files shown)


config.xml
```diff
@@ -11,6 +11,8 @@
     <plugin name="cordova-plugin-whitelist" spec="1" />
     <plugin name="cordova-plugin-code-push" version="1.8.2-beta" />
     <plugin name="cordova-plugin-splashscreen" version="3.2.2" />
+    <plugin name="cordova-plugin-keyboard" version="1.1.4" />
```
nuruddeensalihu: Added this because is thinking of using to address the keyboard popup issue.

config.xml
```diff
@@ -20,6 +22,8 @@
     <allow-intent href="mailto:*" />
     <allow-intent href="geo:*" />
     <preference name="BackgroundColor" value="#2a3052" />
+    <preference name="KeyboardShrinksView" value="true" /> 
+    <preference name="orientation" value="portrait" />
```
nuruddeensalihu: Set the orientation to portrait temporarily .

src/_data/Auth.js
```diff
@@ -30,13 +30,47 @@ export const signout = (nextState, replace) => {
...
     const authorized = store.getState().appState.get('authorized');
     if (!authorized) {
         const oAuthUrl = `https://oauth.binary.com/oauth2/authorize?app_id=${window.BinaryBoot.appId}`;
-        window.open(oAuthUrl);
-        // window.location = oAuthUrl;
+        const winAuth = window.open(oAuthUrl, '_blank', 'location=no');
+        winAuth.addEventListener('loadstop', (e) => {
+             if (e.url.indexOf('acct1') > -1) {
+                 parseUrl(e.url);
+                 winAuth.close();
```
nuruddeensalihu: A reload should take place here before `winAuth.close();` its a challenge at the moment because our device is not running on a server. Would find a way to by tomorrow.

www/home.html
```diff
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html>
 <head>
-    <meta http-equiv="Content-Security-Policy" content="default-src * 'self' data: gap: https://ssl.gstatic.com 'unsafe-eval'; style-src 'self' 'unsafe-inline'; media-src *">
```
nuruddeensalihu: Content seucrity policy is correct. Hwoever, would modifiy to each according to what it does. Example img-src for image links instead of given them access from the content source. This however is optional as our current policy should be ok.

www/home.html
```diff
@@ -30,8 +30,8 @@
 <div id="root" class="root app">
     <div class="mobile-page">
         <img class="spinner" src="img/binary-symbol-logo.svg" />
-        <p class="event listening">Connecting to Device</p>
-        <p class="event received">Device is Ready</p>
+        <p class="listening">Connecting to Device</p>
```
borisyankov: Do we need these?
I don't think so. 
Also need translation if we needed them.

nuruddeensalihu: No we can remove them. Not needed at all. Just optional.

PR #389, opened 2016-07-05
Qingwei/fix mobile login back button

qingweibinary:


qingweibinary:
OK to merge, works properly in Android

PR #399, opened 2016-07-12
Qingwei/mobile perf

qingweibinary:


qingweibinary:
Do not merge

src/trade-params/TradeParams.js
```diff
@@ -93,7 +93,12 @@ export default class TradeParams extends Component {
         this.onAssetChange();
     }
 
-    shouldComponentUpdate = shouldPureComponentUpdate;
+    shouldComponentUpdate(nextProps, nextState) {
+        if (nextProps.disabled && this.props.disabled) {
```
borisyankov: Is that really helping?
If the other diff is correct it shouldn't.

qingweibinary: The intention is not what you think, disabled is equal, but other props can be different, but it would be meaningless to render those changes for Params, since it's disable, and those changes could be re-render all at once, right after disabled is false

but not working well, I'll revert this

qingweibinary:
OK to merge

PR #402, opened 2016-07-12
Nuru/fix statusbar

DanSallau:
Hi,

Attached contained few changes that should add support for status bar plugin for our device. If you don't want to use plugin, you can add the following line in MainViewController.m . probably inside ios platform folder.

```

//fix not hide status on ios7
- (BOOL)prefersStatusBarHidden
{
    return YES;
}
```

The plugin offer many option though. You can hide, show, change color, modify the statusbar as described here. https://github.com/apache/cordova-plugin-statusbar . If you choose the first solution, you can discard the index.html, config.xml changes . Either of the would work. I tested.

### DIFF
--- a/www/index.html
+++ b/www/index.html
@@ -62,6 +62,9 @@
             },
             onDeviceReady: function() {
                 document.addEventListener("backbutton", onBackPressed, false);
+                if (cordova.platformId.toLowerCase() === 'ios') {
+                    StatusBar.hide();
+                }
                 codePush.sync();
             },
             // onDeviceResume: function () {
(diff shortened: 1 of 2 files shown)


www/index.html
```diff
@@ -62,6 +62,9 @@
             },
             onDeviceReady: function() {
                 document.addEventListener("backbutton", onBackPressed, false);
+                if (cordova.platformId.toLowerCase() === 'ios') {
```
borisyankov: That is totally not what we want.
We want it not to overlay with main UI, but not hidden.

PR #405, opened 2016-07-13
Qingwei/temp fix resubscribe issue

qingweibinary:


qingweibinary:
OK to merge

borisyankov:
Temp fix? Does not sound inspiring? :)
Why temporary?

qingweibinary:
@borisyankov because backend should fix the resubscribe issue, this is more of a workaround to reduce it ?

PR #412, opened 2016-07-14
Qingwei/debounce user input

qingweibinary:
@borisyankov I suggest deploy this to beta link 1st


borisyankov:
Looks correct enough. Lets deploy to production :)

PR #413, opened 2016-07-14
Add dev build for mobile

qingweibinary:


package.json
```diff
@@ -16,6 +16,7 @@
     "start": "node server.js",
     "clean": "rimraf dist",
     "build": "npm run clean && npm run build:webpack",
+    "build:dev": "webpack --config webpack.config.js",
```
borisyankov: How do you use that?

qingweibinary: it will build an unminify app.js

qingweibinary: There's still quite a few ways to improve build time, but I have yet tried, something related to DLL

borisyankov: If we don't include eslint in the build, it is almost 2x faster.

There is also some way to tell webpack to skip processing some library files, it was drastic improvement but was not working ;) Needed more work.

Also another way is there some parallel build, haven't examined it.

Not sure the compact thing works, maybe I did't configure it correctly, since the speeds look the same, if you can figure out what I did wrong, would be great.

qingweibinary: I tried parallel build, provided by happypack, only around 10secs improvement, deduplication and minifying are taking half the time

And there's some analyse tool from webpack site which can tell you what to optimize, I've yet understand though

Right now this is the cheapest way to gain the speed

borisyankov: There is a better analyse tool than the default one.
Next we should split into several files and not rebuild the lib ones, highcharts change once in a month or even less.

PR #415, opened 2016-07-14
Nuru/fix setting address

DanSallau:
Hi,

Attached contained little validations that i think should enhance our SettingsAddress page as its done in binary.com for your perusal.


borisyankov:
Examine how the validation and server handing is done in virtual account creation.
Many issues with this approach - non translated errors etc.

nuruddeensalihu:
@boris have thought as well also. But the problem is the server errors with regard to the address and the phone number are not specific. SO i think we have to use the front end . Example for address it returns something like. `Input validation failed: address_line_1`. In the case of the phone number length , the validation actually takes place in the front end. There is no such validation in the backend. But considering the binary.com applies the same validation which i think is a good idea btw, i think that's why they are asking about it.

borisyankov:
I wasn't that much against the validation itself, than the way we store it.
Examine in detail the virtual account creation and most of the new Settings pages.
We do not store the errors in state but calculate them each time, we have a validatedOnce prop that is set to true on first try, we have onSubmit for the form etc.
Just follow the same approach.

nuruddeensalihu:
@borisyankov  have moved the error to the jsx/return method as suggested . One more thing to remember. Unlike the create account, here we do not have a server validation for phone number length. This feature is solely a front end validation on the binary.com itself.

src/settings/SettingsAddress.js
```diff
@@ -36,22 +38,47 @@ export default class SettingsAddress extends PureComponent {
...
+			phone,
+		};
+		if (!is_an_update) {
+			return;
+		}
+
+		if (phone.length < 6) {
+			this.setState({ phoneError: 'length' });
+		} else if (phone.match(/[a-z]/i)) {
+			this.setState({ phoneError: 'allowed' });
```
qingweibinary: why set error to `allowed` ? or you actually want to remove error?

PR #416, opened 2016-07-14
Add six monts to the default exclude date

DanSallau:
Hi,

Set default date value for self exclusion exclude from feature.

Note: Haven't tested well.

### DIFF
--- a/src/settings/SettingsSelfExclusion.js
+++ b/src/settings/SettingsSelfExclusion.js
@@ -4,7 +4,8 @@ import InputGroup from 'binary-components/lib/InputGroup';
 import showError from 'binary-utils/lib/showError';
 import showInfo from 'binary-utils/lib/showInfo';
 import * as LiveData from '../_data/LiveData';
-import monthsAfterStr from 'binary-utils/lib/monthsAfterStr';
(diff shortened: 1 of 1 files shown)


qingweibinary:
@nuruddeensalihu test failed, npm is quite instable recently

nuruddeensalihu:
@qingweibinary i referenced a yet to exist function. That's why.

src/settings/SettingsSelfExclusion.js
```diff
@@ -34,7 +36,8 @@ export default class SettingsSelfExclusion extends Component {
 			max_30day_losses: props.max_30day_losses,
 			max_open_bets: props.max_open_bets,
 			session_duration_limit: props.session_duration_limit,
-			timeout_until: props.timeout_until,
+			timeout_until_time: props.timeout_until,
```
qingweibinary: why suddenly we are using underscore, instead of camelCase?

borisyankov: This is for API call.

borisyankov:
@nuruddeensalihu Did you use the latest name of the call?

PR #419, opened 2016-07-15
Do not allow JP client to use next-gen

qingweibinary:


src/_data/LiveData.js
```diff
@@ -58,6 +60,12 @@ export const changeLanguage = langCode => {
 };
 
 const initAuthorized = async (authData, store) => {
+    if (/japan/.test(authData.authorize.landing_company_name)) {
```
borisyankov: We can just hide the trade panel for now.
Why not allow them to use all the rest?

PR #421, opened 2016-07-18
Qingwei/faster user input

qingweibinary:


www/index.html
```diff
@@ -67,7 +67,7 @@
             },
             onDeviceReady: function() {
                 document.addEventListener("backbutton", onBackPressed, false);
-                codePush.sync();
+//                codePush.sync();
```
borisyankov: Hmmm?

qingweibinary: Want to suggest we disable it at the moment? seems not reliable, and we have not verify it, it caused problem with QA as they keep test on an outdated version despite you've published

borisyankov: The moment we publish a line with it, this version becomes non auto-updatable.

borisyankov: We better debug that as a first priority.

qingweibinary: Added it back

PR #422, opened 2016-07-18
Qingwei/global user input lock

qingweibinary:


qingweibinary:
@borisyankov OK to merge

I prevent user input to accumulate by applying a half transparent div right after user make purchase, so any user tap after that will be ignore, until purchase call finished

src/loading-view/TransparentDiv.js
```diff
@@ -0,0 +1,19 @@
+import React, { PureComponent } from 'react';
+
+export default class TransparentDiv extends PureComponent {
+    render() {
+        return (
+            <div
+                style={{
+                    height: '100vh',
```
borisyankov: That is literally a css and a < div className="invisible" > or something ;)

qingweibinary: Amended =)

Gulpfile.js
```diff
@@ -9,7 +9,7 @@ const sass = require('gulp-sass');
 // const zip = require('gulp-vinyl-zip');
 
 const files = {
-    dist: './dist',
+    dist: './www',
```
borisyankov: Hmmm. Not sure if correct?

qingweibinary: You're right, I'll close this

PR #423, opened 2016-07-19
Qingwei/number comman mobile

qingweibinary:


src/_data/LiveData.js
```diff
@@ -60,7 +59,7 @@ export const changeLanguage = langCode => {
 
 const initAuthorized = async (authData, store) => {
     if (/japan/.test(authData.authorize.landing_company_name)) {
-        showError('Sorry, for japan user please login through www.binary.com ');
+        alert('Sorry, for japan user please login through www.binary.com '); // TODO: use showError without breaking test
```
borisyankov: Ah :)

qingweibinary: This is breaking test, we need to find a way to fix it, but now I think having CI is more important

borisyankov: Yeah, I'll fix it before reverting to showError. Didn't know why you did change it the first time ;)

PR #425, opened 2016-07-20
Downgrade react to 15.1

qingweibinary:


borisyankov:
Did you report the bug?
Isn't this an overreaction?

PR #426, opened 2016-07-20
disable fulscreen

DanSallau:
Before i forget, we should disable the fullscreen mode. It does not good to us and its affecting the keyboard on android.

### DIFF
--- a/config.xml
+++ b/config.xml
@@ -22,7 +22,7 @@
     <allow-intent href="mailto:*" />
     <allow-intent href="geo:*" />
     <preference name="android-minSdkVersion" value="19"/>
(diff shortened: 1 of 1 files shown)


config.xml
```diff
@@ -22,7 +22,7 @@
     <allow-intent href="mailto:*" />
     <allow-intent href="geo:*" />
     <preference name="android-minSdkVersion" value="19"/>
-    <preference name="Fullscreen" value="true"/>
+    <preference name="Fullscreen" value="false"/>
```
borisyankov: You can remove the setting completely, if we decide to do so.

PR #429, opened 2016-07-21
Qingwei/fix nan on purchase

qingweibinary:


src/_actions/TradeActions.js
```diff
@@ -171,6 +171,17 @@ export const updatePurchaseInfo = (index, fieldName, fieldValue) => ({
...
+// TODO: this is arbritary, need discuss how we should handle it
+const trySubscribeUntilSuccess = c =>
+    LiveData.api
+        .subscribeToOpenContract(c)
+        .then(response => {
+            const contract = response.proposal_open_contract;
+            if (contract.validation_error && Object.keys(contract).length < 3) {
+                setTimeout(() => trySubscribeUntilSuccess(c), 300);
+            }
+        });
```
qingweibinary: @borisyankov should we even retry ?
Backend might fail for all kind of reason that frontend does not have information about, do you think keep on retry make sense?

src/contract-details/ContractReceipt.js
```diff
@@ -15,23 +15,30 @@ export default class ContractReceipt extends PureComponent {
 
 	render() {
 		const { contract, showLongcode, onTradeAgainClicked } = this.props;
+		const backendFailure = contract.validation_error && Object.keys(contract).length < 3;
```
borisyankov: Is this a hack or a documented thing?
What significance has the 'contract keys less than 3'

src/_actions/TradeActions.js
borisyankov: Nope. Not at all.

src/contract-details/ContractReceipt.js
qingweibinary: This is pretty random, failure is indicated when the response contains `contract_id` and `validation_error` only.

I submitted my suggestion in v4 suggestion, and add this change here.

src/_actions/TradeActions.js
qingweibinary: ok, will remove

qingweibinary: Take note this would mean there's a chance, purchased contract will not be shown in our portfolio, unless user refresh

Should be edge case, reported by binary-tester

src/_actions/WorkspaceActions.js
```diff
@@ -49,8 +49,9 @@ export const changeActiveLayout = (tradesCount, layoutN) =>
             combinedAssetChoices.forEach(asset => dispatch(selectAsset(asset)));
             return dispatch(updateActiveLayout(tradesCount, layoutN, combinedAssetChoices.toJS()));
         }
-        otherAssetInSameMarket.forEach(asset => dispatch(selectAsset(asset)));
-        return dispatch(updateActiveLayout(tradesCount, layoutN, otherAssetInSameMarket.toJS()));
+        const additionAsset = otherAssetInSameMarket.take(additionTradeNeeded);
```
borisyankov: What does additionAsset mean?

src/portfolio/PortfolioItem.js
```diff
@@ -16,7 +16,8 @@ export default class PortfolioItem extends PureComponent {
 
     render() {
         const { contract } = this.props;
-
+        const backendFailure = contract.validation_error && Object.keys(contract).length < 3;
```
borisyankov: Extract this logic into a backendFailure = contract => thing and reuse.
Add a single line of comment that this is hackish.
Ideally will remove when backend starts behaving.

src/_actions/WorkspaceActions.js
qingweibinary: Asset to add, I'll rename it

qingweibinary:
Fixed as comments

PR #431, opened 2016-07-21
Replace intl polyfill with an npm install

DanSallau:
Hi,

Attached is what works for me on ipod devices. :wink:

### DIFF
--- a/src/_store/BootProvider.js
+++ b/src/_store/BootProvider.js
@@ -1,5 +1,6 @@
 import React, { Children, PureComponent, PropTypes } from 'react';
 import { connect } from 'react-redux';
+import Intl from 'intl';
 import { IntlProvider, addLocaleData } from 'react-intl';
 import allTexts from '../_constants/texts';
 import locale from '../_constants/languageLocaleMap';
--- a/www/index.html
+++ b/www/index.html
@@ -78,7 +78,6 @@
 </script>
 
 <script src="boot.js"></script>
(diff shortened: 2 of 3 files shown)


src/_store/BootProvider.js
```diff
@@ -1,5 +1,6 @@
 import React, { Children, PureComponent, PropTypes } from 'react';
 import { connect } from 'react-redux';
+import Intl from 'intl';
```
borisyankov: How large is it?
It might not be worth statically increasing our bundle size just for iPod.
Also this can be reported as bug to polyfill.io and they will likely fix it.

www/index.html
```diff
@@ -78,7 +78,6 @@
 </script>
 
 <script src="boot.js"></script>
-<script src="https://cdn.polyfill.io/v2/polyfill.min.js?features=fetch,Promise,Array.from,Array.prototype.find,Array.prototype.includes,String.prototype.startsWith,String.prototype.includes,Math.sign,Intl.~locale.en,Intl.~locale.id,Intl.~locale.ru,Intl.~locale.fr&unknown=polyfill"></script>
```
borisyankov: Noooooooope.
This includes many other polyfills.

borisyankov: Just including everything statically is not the solution.

nuruddeensalihu: OH have actually opened an issue with them . I Would update this when they reply. Nonetheless the size is not that much . Barely noticeable.

borisyankov: You need to include also all the locales. Which are not that small.
Also need to include all polyfills, fetch, Promise, Array.from, Array.find etc. There are a dozen of them. Which is large.

borisyankov: I don't think iPod is that important to bloat the code that much. We better wait for a fix from polyfill.io

PR #432, opened 2016-07-22
Qingwei/consolidate api reduce render

qingweibinary:
OK to merge


src/_actions/TradeActions.js
```diff
@@ -82,81 +82,85 @@ export const updateTradeError = (index, errorID, error) => ({
...
-            stopType,
-            stopProfit,
-            stopLoss,
-            barrierType,
-        } = tradeParam;
-
-        if (!(amount && basis && type && symbol)) {
-            return;
-        }
+const subscriptionQueue = [
```
qingweibinary: @borisyankov this is not very clean imo, the objective is to prevent race condition when user tapping on UI that's faster than network call.

The original problem can be summarize as
1. We fire forget, and subscribe
2. Before api return, user click something, thus we fire forget and subscribe again
3. This would means our 2nd `forget` is based on an old subscription id, but we are supposed to forget the subscription made in `1`
4. Subscribe call in `1` returns, and in a shortwhile subscribe in `2` return and overwrite the `subscribe1`
5. Now we have a phantom subscription, which we do not track thus cannot forget

This is an edge case, not sure if it worth the complexity added.
And do you have cleaner solution in mind?

borisyankov: Good that you are taking this into consideration.
Could be too much of an edge case, and not happen really in practice (or it could).

borisyankov:
Looks good.

src/_actions/TradeActions.js
qingweibinary: OK, I'll remove it and see if people complains

qingweibinary:
Reopen to trigger test

qingweibinary:
@borisyankov I've removed the hacky code that force sequential subscription call, ok to merge from me

PR #434, opened 2016-07-22
Force tradecard to update when duration unit and trade type changes

qingweibinary:


borisyankov:
Reason?

qingweibinary:
@borisyankov as we are now using uncontrolled input, in some cases, the internal state and the UI representation will be out of sync, so we have to force the ui to update.

There are 2 cases 
1. When user change duration unit, this would affect barriers, as intraday contract uses relative barrier, and absolute barrier is used in other cases.

2. Trade type changes will affect every other trade parameters too.

PR #435, opened 2016-07-22
Change input type from number to use text

DanSallau:
Hi,

The issue with our number type text input as reported here https://trello.com/c/KNEHsIXh/702-delete-decimal-number-required-2nd-time-selection does not seem to have a clear solution . Another way we can solve it is by using a text type input instead. Attached contained few changes that should allow us use text input instead for your perusal. I would update the binary-components should this be satisfactory.


borisyankov:
That is not a clear improvement.
Text inputs result in text input keyboards on mobile.
For mobile we should experiment with type=number and type=tel (since for phone numbers the keyboard is better)

nuruddeensalihu:
@borisyankov we could use `inputmode` e.g `inputmode="numeric"` for our text. input type `tel` may not permit a decimal places . Have thought a well for alternative but doesn't seem to find any better. The problem would persist so as far we use `number` input type. Couldn't find a better solution or reason though.

PR #439, opened 2016-07-27
Set default color for root class

DanSallau:
Hi,

When i fullscreen, firefox tend to change set black as the default color of the body hence the need for setting the default color externally.

### DIFF
--- a/styles/_basic.scss
+++ b/styles/_basic.scss
@@ -15,6 +15,7 @@ body {
     width: 100%;
     height: 100%;
     display: flex;
+    background-color: white;
 }
 
 .disabled {


borisyankov:
When does that happen? Sounds weird and I haven't seen that.

nuruddeensalihu:
@borisyankov use the full screen button in firefox

borisyankov:
I did use the full screen button in Firefox (and other browsers). Nothing out of the ordinary.

nuruddeensalihu:
@borisyankov  it does happen if u use the fullscreen button in Firefox thats for sure. Try opening the news tab from the right list and then go fullscreen. There are explanations on that online with regard to FF.

PR #442, opened 2016-07-27
Retrieve countrly list from server

DanSallau:
Get residence list from server instead.


src/_data/LiveData.js
```diff
@@ -133,6 +134,6 @@ export const connect = async store => {
         api.events.on(key, data => store.dispatch(action(data)));
         api.events.on(key, () => window.console.warn);
     });
-
+    api.getResidences();
```
qingweibinary: This is not the place man ~~ 
See where we place such calls, in the same file

nuruddeensalihu: @qingweibinary that call either belong there , or veryfEmailCard to my understanding. "See where we place such calls" like ?

qingweibinary: Sorry, we've removed the initUnAuth, my bad

I think we should call when needed

borisyankov: @nuruddeensalihu I agree with @qingweibinary 
Please do the changes before me merging this.

nuruddeensalihu: @borisyankov where do we need to change ? I think if you agree with qingwei then you agree with me by extension. 😉

qingweibinary: `we should call when needed` ;)

borisyankov:
Also fix merge conflict ;)

PR #444, opened 2016-07-28
Qingwei/refactor to saga

qingweibinary:


qingweibinary:
Super high level view

screen shot 2016-07-28 at 3 16 52 pm

qingweibinary:
## TODO
1. Add subscription call for all params change saga
2. Properly disable purchase button, might need to shift debounce into Saga, using delay
3. Make sure we always sync params change from UI into redux store, although UI does not always reflect from redux store

src/duration-picker/ForwardStartingOptions.js
```diff
@@ -32,72 +28,50 @@ export default class ForwardStartingOptions extends PureComponent {
...
-        const allowStartLater = !!forwardStartingDuration;
-        const onlyStartLater = allowStartLater && !options;
 
         const defaultDate = new Date(defaultDateStart * 1000);
         const defaultTime = epochToUTCTimeString(defaultDateStart);
-
         return (
             <div className="param-row forward-starting-picker">
-                {allowStartLater && <Label text={'Start Time'} />}
+                {<Label text="Start Time" />}
```
borisyankov: You don't need the {} now ;)

borisyankov:
After carefully reviewing the whole code, all I can say is: hard to judge ;)

Keep writing tests, if you think on Tuesday it is good enough to merge, lets merge and I'll test too and help with unit tests and refactoring, then if we don't find issues on Wed we deploy to app.binary.com/beta for testing.

Sounds reasonable?

qingweibinary:
Yes ~

PR #445, opened 2016-07-28
Fix start later time issue in FF and safari

DanSallau:
Hi, 

Attached contain little changes that should address the card here for your perusal. 
https://trello.com/c/jLKYQwhV/765-firefox-and-safari-parameters-of-the-trading-will-be-reset-if-user-modify-the-start-time

### DIFF
--- a/src/duration-picker/ForwardStartingOptions.js
+++ b/src/duration-picker/ForwardStartingOptions.js
@@ -1,6 +1,6 @@
 import React, { PureComponent, PropTypes } from 'react';
 import { epochToUTCTimeString, dateToEpoch,
-    timeStringToSeconds, dateToDateString, returnValidaDate, returnValidTime } from 'binary-utils';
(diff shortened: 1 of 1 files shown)


src/duration-picker/ForwardStartingOptions.js
```diff
@@ -42,7 +42,9 @@ export default class ForwardStartingOptions extends PureComponent {
 
     onTimeChange = e => {
         const { dateStart } = this.props;
-        const inputValue = e.target.value;
+        const inputValue = e.target.value.split(':')
```
borisyankov: This looks to me like something worthy of separate function + unit tests.

nuruddeensalihu: ok, but we are not using it anywhere and not likely to. Its not complex or complicated also. That's why i didn't think of the necessity.

nuruddeensalihu:
@borisyankov  have moved the validation to binary-utils here https://github.com/binary-com/binary-utils/pull/29

PR #446, opened 2016-07-28
Qingwei/block until active symbols

qingweibinary:


borisyankov:
You mentioned, this is not needed with the saga refactor?

qingweibinary:
@borisyankov yep, can be closed

PR #457, opened 2016-08-02
Warn user when a wrong time fromat is used

DanSallau:
The attached PR show simply warn user on the timeformat should the format be wrong . E.g `33:333:23` . Hrs shouldn't be greater than 24 presumably Hence the warning.  This however is not a wrong time as 33 hrs can be translated to 2 days 9 hrs and the two days be added to the date field should the user choose to proceed. 

Ticket : https://trello.com/c/kv3hj0KG/37-github-nuru-safari-firefox-no-validations-for-date-time-fields


src/duration-picker/ForwardStartingOptions.js
```diff
@@ -43,6 +43,9 @@ export default class ForwardStartingOptions extends PureComponent {
     onTimeChange = e => {
         const { dateStart } = this.props;
         const inputValue = returnValidTime(e.target.value, ':');
+        if (isValidTime(inputValue)) {
+             showError('Time format is wrong!');
```
borisyankov: We are not going to use showError to display errors any more.

nuruddeensalihu: @borisyankov  what are we going to use instead 
?

borisyankov: ErrorMsg component

nuruddeensalihu: @borisyankov i actually wanted to avoid that even though it would better though in this case. T o use the component, we might have to use the component state object.

src/duration-picker/ForwardStartingOptions.js
```diff
@@ -43,6 +44,7 @@ export default class ForwardStartingOptions extends PureComponent {
     onTimeChange = e => {
         const { dateStart } = this.props;
         const inputValue = returnValidTime(e.target.value, ':');
+        this.setState({ isInValidTime: isValidTime(inputValue) });
```
borisyankov: We don't need this to be state, in can be calculated on render.

src/trade-params/saga/DurationSaga.js
```diff
@@ -142,7 +142,9 @@ export function* handleStartTimeChange(action) {
     const contractPerType = contractNeeded[tradeCategory][type];
     const durationAllowed = allTimeRelatedFieldValid(newDateStart, duration, durationUnit, contractPerType);
 
-    if (durationAllowed) {
+     if (isValidTime(time)) {
+        yield put(updateTradeError(index, 'durationError', 'Time format is wrong!'));
```
qingweibinary: Note that yield is not return, if you want to exit early, explicit `return` is needed

@nuruddeensalihu

nuruddeensalihu: I think either of the two serve our purpose here, There isn't a code block after the if statement so the block exit there. Plus 'return' is part of what the yield keyword does it self if i understand correctly. I think either of the two should be fine(i.e return or not ). https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Operators/yield

qingweibinary: It's better to have

```
if (xxx) {
   ...
   return
}

if (yyyy) {
   ...
   return
}
```

than to have

```
if (xxx) {
   ...
   return
} else if (yyyy) {
   ...
   return
}
...
```

The first one is generally better, it tells reader that code exit with that condition, and that's all it does, 2nd chunk however requires reader to go until the end of code to know what's going on

I might have make the mistake at the first place, would you help me to fix it altogether?
Thanks

nuruddeensalihu: @qingweibinary would have a look.

PR #464, opened 2016-08-04
Fix code to adapt new contract rules, add test to verify, fix bug cau…

qingweibinary:
…sed by backend changes


qingweibinary:
It's failing as it depends on https://github.com/borisyankov/binary-test-data/pull/4

PR #480, opened 2016-08-15
Remove assets from cache, fix issue when switching from mlt to mf

qingweibinary:


qingweibinary:
Note: This would increase loading time

PR #486, opened 2016-08-17
Qingwei/improve chart

qingweibinary:


src/_trade/validation/isBarrierCountValid.js
```diff
@@ -5,11 +5,11 @@ export default (barrier, expiryType, contractPerType) => {
     let b1 = true;
     let b2 = true;
     if (barrierOptions[0]) {
-        b1 = barrier[0];
+        b1 = barrier[0] || barrier[0] === 0;
```
borisyankov: Hmmm, what does that line mean?

qingweibinary: zero is falsy but valid barrier value ~

I'll refactor it so that it's less cryptic

PR #487, opened 2016-08-17
Qingwei/fix react bug

qingweibinary:


qingweibinary:
This PR depends on https://github.com/binary-com/binary-components/pull/16

PR #489, opened 2016-08-18
Qingwei/trackjs

qingweibinary:


www/index.html
```diff
@@ -64,7 +64,7 @@
                     navigator.app.exitApp();
                 }
             }
-            if (window.location.hash === '#/') {
+            if (window.location.hash.startsWith('#/?_')) {
```
borisyankov: I thought you will use another method?

qingweibinary: it does not works, remember I mentioned `pathname` gives the whole path in mobile

any alternative?

PR #494, opened 2016-08-19
Fix exclude account login

DanSallau:
Hi,

Attached few changes that should allow the Virtual account to be used incase of exclusion as mentioned here https://trello.com/c/hk1QgRpg/44-self-exclusion-behaviour .


src/_store/root.js
```diff
@@ -35,7 +35,15 @@ export default class Root extends PureComponent {
         try {
             await tryAuth(token);
         } catch (e) {
-            actions.updateAppState('authorized', false);
+            await actions.updateAppState('authorized', false);
+            const errorMsg = e.message;
+            if (errorMsg.indexOf('you have excluded yourself until') > -1) {
```
qingweibinary: This is probably wrong, will not work on translated message, or whenever backends change the msg, use error code?

src/settings/SettingsSelfExclusion.js
```diff
@@ -68,13 +68,19 @@ export default class SettingsSelfExclusion extends PureComponent {
...
 		};
+
 		try {
 			await api.setSelfExclusion(newSelfExclusionSettings);
 			this.setState({
 				success: true,
 				serverError: false,
 			});
-			setTimeout(() => this.setState({ success: false }), 3000);
+			setTimeout(() => {
```
qingweibinary: Why is a setTimeout appear here?

src/_store/root.js
qingweibinary: I prefer to handle this in tryAuth itself, as a recursive call, what do you think?

borisyankov: Likely need to catch by error type, since you will not have this string when translated :)

src/settings/SettingsSelfExclusion.js
nuruddeensalihu: @qingweibinary would actually changed it by updating the live-api call as mentioned on the another on trello. I am thinking of making the call returns with api error response from the liveapi instead of throwing from live api. So the message here was just temporary pending the changes from live-api. I would use ErrorCode or something similar.

borisyankov: Not sure I am clear on what you intend to do.

nuruddeensalihu: @borisyankov  this is what i am thinking. Please see the update and here https://github.com/binary-com/binary-live-api/pull/70/files

src/_data/Auth.js
```diff
@@ -6,7 +6,14 @@ import { updateAppState, removePersonalData, updateToken, updateBoot } from '../
 
 export const tryAuth = async token => {
     store.dispatch(updateAppState('authorized', false));
-    const response = await api.authorize(token);
+    let response = await api.authorize(token);
+    if (response.error && response.error.code === 'SelfExclusion') {
+          const account = window.BinaryBoot.accounts.filter(x => x.account.startsWith('VRTC'));
```
qingweibinary: I think `find` is more appropriate

src/settings/SettingsSelfExclusion.js
qingweibinary: @nuruddeensalihu you mention would actually change it?

src/_data/Auth.js
borisyankov: Also, there is no guarantee that there will be a VRTC account!

nuruddeensalihu: @borisyankov i think there would always be Virtual account wouldn't there ?

borisyankov: Nope.

qingweibinary:
@nuruddeensalihu this have been here for 11 days, last activity is 6 days ago

What's stopping it from being merge?

nuruddeensalihu:
@qingweibinary I don't see any issue with this PR, though boris said there are instance where VRTC account would not exist . If indeed there is, then there would be issue then. I mean could a user have only MF account without VRTC ? IF so then this solution would not work for such case. Otherwise this PR is ok. I have tested many times and no problem with it.

qingweibinary:
@nuruddeensalihu I think it's pretty clear that boris said that VRTC does not always exists.

1. Legacy account does not have virtual account.
2. whether we agree or not, it's important to reply, so that we can proceed instead of hanging the PR somewhere.

So I guess you will make some changes so that it handle cases where VRTC is absent, right ?

nuruddeensalihu:
@qingweibinary i am out of idea when he said VRTC does not exist in some instance. Thats something i have never thought of. So i was thinking if indeed it happened randomly that some account does not have VRTC then this PR would be obsolete. Btw what kind of account do legacy account use in place of VRTC ? I mean do they have an alternative or just no VRTC at all?

qingweibinary:
@nuruddeensalihu 

if I remember correctly, there is no other account.

PR #501, opened 2016-08-23
Replace tick stream with tick history

DanSallau:
Replace tick stream with tick history for watchlist.

https://trello.com/c/CQUfIUva/2-handle-the-exception-this-market-is-presently-closed


borisyankov:
@qingweibinary Review please?

src/_actions/WatchlistActions.js
```diff
@@ -3,9 +3,13 @@ import { WATCHLIST_TOGGLE_ASSET } from '../_constants/ActionTypes';
 import { api } from '../_data/LiveData';
 
 export const watchlistToggleAsset = (symbol, isSubscribed) =>
-    dispatch => {
+    async (dispatch) => {
         if (isSubscribed) {
-            api.subscribeToTick(symbol);
+            try {
+                await api.getTickHistory(symbol, { end: 'latest', count: 10, subscribe: 1 });
```
qingweibinary: we probably need `adjust_start_time: 1`

src/_data/LiveData.js
```diff
@@ -84,13 +84,19 @@ const initAuthorized = async (authData, store) => {
             }
         });
 
-
+    const susbcribeWatchlist = async (symbol) => {
```
qingweibinary: bad naming, `subscribeWatchList` and `subscribeToWatchList` are too similar

src/_data/LiveData.js
```diff
@@ -84,13 +84,19 @@ const initAuthorized = async (authData, store) => {
...
+        try {
+            await api.getTickHistory(symbol, { end: 'latest', count: 10, subscribe: 1 });
```
qingweibinary: Also I am guessing we need `adjust_start_time: 1`

src/_data/LiveData.js
qingweibinary: And this method have nothing to do with watchlist, I think an inline closure would do

src/_data/LiveData.js
```diff
@@ -43,6 +43,13 @@ export const changeLanguage = langCode => {
     api.getTradingTimes(new Date());
 };
 
+export const getTickHistory = async (symbol) => {
```
qingweibinary: this export looks really random, how about this?

we place this logic in action creator, then dispatch action from LiveData, which is the common pattern we have been using.

nuruddeensalihu: @qingweibinary  i don't think thats necessary. The method does change a state in any way beside making a call to an WS. HAd it been we are updating or changing a state example let's say logging history or something then would have made sense. I think it should be ok like this just like how we handle the one above it.

qingweibinary: OK

src/_data/LiveData.js
```diff
@@ -43,6 +43,13 @@ export const changeLanguage = langCode => {
...
+    try {
+        await api.getTickHistory(symbol, { end: 'latest', count: 10, subscribe: 1, adjust_start_time: 1 });
+    } catch (err) {
+        await api.getTickHistory(symbol, { end: 'latest', count: 10 });
```
borisyankov: I'm much more in favor of using the subscribe separately.
Then it will be:

```
getTickHistory()
try {
  subscribeToTicks();
catcth() {
 // suppress error, do nothing
}

qingweibinary: I think both has it's merit, eg. use only tickHistory subscribe save round trip

So no comment on this.

@nuruddeensalihu fix the test and push boris to merge ?

borisyankov: OK, agree this this is ok way to do it, though I am generally not liking the subscribe: 1 concept in the api even :)

PR #505, opened 2016-08-23
Qingwei/disable range selector in purchase

qingweibinary:


src/_reducers/TickReducer.js
```diff
@@ -17,7 +17,7 @@ export const mergeTicks = (existingTicks, newTicks) => {
 
     const lastNewTicksEpoch = getLast(newTicks).epoch;
     const oldestExistingTickEpoch = existingTicks[0].epoch;
-    const lastExistingTickEpoch = getLastTick(existingTicks).epoch;
+    const lastExistingTickEpoch = getLast(existingTicks).epoch;
```
borisyankov: BTW getLast turns out is an overkill 💃 
because [][0] is undefined anyways, it is not an error 👯

qingweibinary: it at least serve as a shorthand of

superUglyAndLongVariableName[superUglyAndLongVariableName.length - 1]

PR #506, opened 2016-08-23
fix getLast error

DanSallau:
Little check on our state .


borisyankov:
Good, but when does that happen?
@qingweibinary @nuru We should not miss to add unit-tests for these cases.

qingweibinary:
yes, we should add test, happens when there's something in watchlist , which trigger tick streams.

Nuru:
@borisyankov please leave me out of this. I am not @nuruddeensalihu

PR #508, opened 2016-08-24
Nuru/fix take last

DanSallau:
Fix take last issue with test.


src/_reducers/__tests__/TickReducer-test.js
```diff
@@ -16,7 +17,17 @@ describe('TickReducer', () => {
             expect(actual.get('x').toJS()).to.have.lengthOf(1);
         });
     });
-
+    describe('should not throw', () => {
```
borisyankov: You need a better explanation.
Should not throw in what case?

PR #510, opened 2016-08-24
Nuru/android release

DanSallau:
Hi

Attached contained few changes that should automate our android release for your perusal.


Gulpfile.js
```diff
@@ -19,6 +21,11 @@ const files = {
     js: './src',
     static: ['./www/**/*', './config.xml', './electron.js', '!./www/**/*.scss'],
     sass: './styles/*.scss',
+    androidApk: './platforms/android/build/outputs/apk',
+    unalignedApk: './platforms/android/build/outputs/apk/android-release-unaligned.apk',
+    alignedApk: './platforms/android/build/outputs/apk/android-release-aligned.apk',
+    zipAligned: '/Users/nuru/Library/Android/sdk/build-tools/23.0.3/zipalign', // Note the path to the zipalign on your pc
+    keyPassword: '****', // make sure you don't push the correct password here to github. Edit password before running task
```
binarytom: Very easy to push the password by mistake - can you change this to read the password from a .gitignore-listed external file or user input instead?

Gulpfile.js
```diff
@@ -19,6 +21,11 @@ const files = {
     js: './src',
     static: ['./www/**/*', './config.xml', './electron.js', '!./www/**/*.scss'],
     sass: './styles/*.scss',
+    androidApk: './platforms/android/build/outputs/apk',
+    unalignedApk: './platforms/android/build/outputs/apk/android-release-unaligned.apk',
+    alignedApk: './platforms/android/build/outputs/apk/android-release-aligned.apk',
+    zipAligned: '/Users/nuru/Library/Android/sdk/build-tools/23.0.3/zipalign', // Note the path to the zipalign on your pc
```
binarytom: Two minor comments:

* zipAligned implies an .apk that has been zip-aligned. This is the tool, so better to call it `zipAlign`, or just `zipalign` to match the actual file name (and update line 125 as well)
* Use the `ANDROID_SDK_ROOT` env var instead of hardcoding the full path (e.g. `require('gulp-util')`, then `gutil.env.ANDROID_SDK_ROOT`)

Gulpfile.js
borisyankov: It is virtually guaranteed to happen once out of 10 times. ;)

Gulpfile.js
nuruddeensalihu: @borisyankov ANDROID_SDK_ROOT var doesn't seem to be available in the gutil . Where exactly do i set it @binarytom ?

nuruddeensalihu: @binarytom inorder to have ANDROID_SDK_ROOT avail in gutil.env , we need to hard code it in package.json don't we? Or is there any other way we can fetch it?

binarytom: @nuruddeensalihu `ANDROID_SDK_ROOT` is an environment variable that should be available after setting up an Android development environment. It points to the top-level SDK directory (so on your system, you could set it with `export ANDROID_SDK_ROOT=/Users/nuru/Library/Android/sdk`). Google don't document it very well, but it's used by several SDK tools - there are various stackoverflow.com posts describing how to set it on different OSes.

You might not need `gutil.env` for this - I think https://nodejs.org/api/process.html#process_process_env would work better.  The idea is that people should be able to do a build without needing to change `Gulpfile.js`.

PR #514, opened 2016-08-29
Qingwei/debug perf

qingweibinary:
@borisyankov 

Profiling on macbook I am using does not show significant difference, although number of render is reduce a whole lot. Time taken and profile chart both are close

each render spent more than 80% of CPU time in binary, maybe that's a better place to optimize

Note: This is for discussion, do not merge unless you're sure ;)

### DIFF
--- a/src/trade/trade-chart/TradeViewChart.js
+++ b/src/trade/trade-chart/TradeViewChart.js
@@ -67,6 +67,16 @@ export default class TradeViewChart extends PureComponent {
         }
     }
 
+    shouldComponentUpdate(nextProps) {
+        const hasNotEq = Object.keys(this.props).some(k => {
+            if (JSON.stringify(this.props[k]) !== JSON.stringify(nextProps[k])) {
+                return true;
+            }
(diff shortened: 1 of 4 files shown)


borisyankov:
> each render spent more than 80% of CPU time in binary

What does that mean? :)

qingweibinary:
@borisyankov typo

in `binary-charts`

borisyankov:
Hmmm...
I do know why we don't see that much improvement ;(

src/trade/trade-chart/TradeViewChart.js
```diff
@@ -67,6 +67,16 @@ export default class TradeViewChart extends PureComponent {
         }
     }
 
+    shouldComponentUpdate(nextProps) {
+        const hasNotEq = Object.keys(this.props).some(k => {
+            if (JSON.stringify(this.props[k]) !== JSON.stringify(nextProps[k])) {
```
borisyankov: This is not fast.

PR #515, opened 2016-08-29
Prioritize Volatile

qingweibinary:


src/asset-picker/AssetPickerList.js
```diff
@@ -11,10 +11,21 @@ export default class AssetPickerList extends PureComponent {
 
 	render() {
 		const { assets } = this.props;
+		const sorted = Object.keys(assets)		// prioritize Volatile assets
+			.sort(a => {
+				if (a === 'Daily Reset Indices') {
```
borisyankov: Would that work when translated?

qingweibinary: You're right, nope

I think need some changes deep in the selector, will do it and resubmit

PR #517, opened 2016-08-30
Add the exclude acct in recursive function

DanSallau:
Hi,

Please attached contained few changes that should address our exluded acct issue for your perusal.

### DIFF
--- a/src/_store/root.js
+++ b/src/_store/root.js
@@ -24,6 +24,25 @@ addLocaleData({
 
 const emptyObject = {};
 
+
+export const accountExclusion = async (token, accts = []) => {
+    const account = window.BinaryBoot.accounts.find(x => x.account.startsWith('VRTC'));
+    accts.push(token);
+    if (account) {
+        const newToken = account.token;
+        await tryAuth(newToken);
+    } else {
+        const accounts = window.BinaryBoot.accounts.filter(x => !accts.includes(x.token));
+        try {
+            await tryAuth(accounts[0].token);
+        } catch (ex) {
+            if (ex.error && ex.error.error.code === 'SelfExclusion') {
+                accountExclusion(accounts[0].token, accts);
+            }
+        }
+    }
+};
+
 export default class Root extends PureComponent {
 
     async componentWillMount() {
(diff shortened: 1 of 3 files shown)


src/_store/root.js
```diff
@@ -24,6 +24,25 @@ addLocaleData({
 
 const emptyObject = {};
 
+
+export const accountExclusion = async (token, accts = []) => {
```
borisyankov: Does that need to be in root.js ?

nuruddeensalihu: Nope, we can take it to utils.

nuruddeensalihu: @borisyankov i think we cannot take it binary-utils. Also i don't think leaving there would do harm, or should i take it to a diff file?

qingweibinary: Can we improve the parameter name?

does token mean token to try next or?
what does accts means and how is it different from window.BinaryBoot.accounts

src/settings/SettingsSelfExclusion.js
```diff
@@ -74,7 +74,12 @@ export default class SettingsSelfExclusion extends PureComponent {
 				success: true,
 				serverError: false,
 			});
-			setTimeout(() => this.setState({ success: false }), 3000);
+			setTimeout(() => {
```
qingweibinary: I never notice this before, why do we need a settimeout here?

borisyankov: Not suru. Maybe @nuruddeensalihu added it initially too?

nuruddeensalihu: @borisyankov i didn't add . The setTimeout there is what you use for splash message in self exclusion. It was added by you . 😉 . Btw have refactored the var names.

qingweibinary:
@borisyankov @nuruddeensalihu 

anymore issue with this PR?

If not we merge it?

I still have question of the 3 seconds timeout though, nuru, do you think we need it? if not how about we remove it.

borisyankov:
Yeah, my only comment now is the setTimeout.

nuruddeensalihu:
@borisyankov @qingweibinary  the setTimeOut is there in our code at the moment. You can verify that. Like i said when you update selfExclusion, it animate the message for a while which disappear after few seconds. I believe the purpose it served there . 😉

qingweibinary:
@nuruddeensalihu I see, that's fine then, thanks

PR #519, opened 2016-08-31
Add check for unexpected bug

qingweibinary:


qingweibinary:
This will help in TrackJS

PR #524, opened 2016-09-02
Qingwei/fix trading time typo

qingweibinary:


borisyankov:
Flow woulda caught that :)

PR #527, opened 2016-09-05
Update electron builder

DanSallau:
Hi,

Attached contained few changes that should update our electron build for your perusal.


package.json
```diff
@@ -158,6 +158,8 @@
     "chai-subset": "^1.3.0",
     "coveralls": "^2.11.12",
     "css-loader": "^0.24.0",
+    "electron": "^1.3.5",
+    "electron-builder": "^6.4.0",
```
borisyankov: Look in build/package.json
We'll have these dependencies there.

nuruddeensalihu: Couldn't find any . I can see `gulp-atom-electron` though. Are you referring to that ?

borisyankov: All dependencies that are required for building, especially electron, are in build, so a developer does not need to do big updates all the tipe

PR #530, opened 2016-09-05
Fix error

qingweibinary:


qingweibinary:
OK to merge after test passed ~

PR #532, opened 2016-09-06
Qingwei/refactor chart

qingweibinary:
Forgot to submit

### DIFF
--- a/styles/_trade-params.scss
+++ b/styles/_trade-params.scss
@@ -185,9 +185,14 @@
 .trade-chart {
     width: 100%;
     height: 100%;
+    display: flex;
+    flex-direction: column;
 
-    & > div {
-        height: 100%;
+    :nth-child(1) {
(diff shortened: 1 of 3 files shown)


styles/_trade-params.scss
```diff
@@ -185,9 +185,14 @@
 .trade-chart {
     width: 100%;
     height: 100%;
+    display: flex;
+    flex-direction: column;
 
-    & > div {
-        height: 100%;
+    :nth-child(1) {
+        flex: 1 1 auto;
```
borisyankov: Why would 1:9 make sense? 💃

qingweibinary: totally random, looks fine from my machine so I stay with it

borisyankov: See my refactor... ;) Anyways later I'll move this styles to the chart itself.

PR #538, opened 2016-09-09
Nuru/fix electron latest

DanSallau:
Another mile stone in the electron build. Enable locatstorage support.


www/boot.js
```diff
@@ -61,7 +61,7 @@
     readConfig();
     parseUrlAndStoreAccountInfo(window.location.href);
     window.BinaryBoot.parseUrl = parseOAuthResponse;
-    window.BinaryBoot.appId = window.cordova ? 1006 : 1001;
+    window.BinaryBoot.appId = window.cordova || window.electron ? 1006 : 1001;
```
borisyankov: Hmm, do we really want the 'mobile' id passed to Electron?

PR #540, opened 2016-09-09
Qingwei/data

qingweibinary:


src/_reducers/TickReducer.js
```diff
@@ -56,7 +49,7 @@ export default (state = initialState, action) => {
                 epoch: +tick.epoch,
                 quote: +tick.quote,
             };
-            return state.update(symbol, List.of(), v => v.takeLast(1000).push(fromJS(newTick)));
+            return state.update(symbol, List.of(), v => v.takeLast(10000).push(fromJS(newTick)));
```
borisyankov: Is 10K the correct amount?

qingweibinary: depends on what you mean by correct ;)

for ticks, I hit 10000 easily, so might raise it or we discuss how to handle

PR #558, opened 2016-09-20
Ff flex

DanSallau:
Hi,

Attached contained readjustments that should perfectly fit our layouts for your perusal. Issues like

image

should be made to perfectly fit the window. You may noticed that the row two is suppose to accommodate 4 trades, it rather accommodate 3 . This should be fixed and fit across all browsers.

Also Firefox performance should increase. I have test severally , the crashing and hanging issue should not occur.

### DIFF
--- a/src/layouts/layouts.css
+++ b/src/layouts/layouts.css
@@ -75,8 +75,8 @@
 }
 
 .layout-2-1 > div {
-    flex: 1;
-    max-height: 50%;
+    flex: 1 1 50%;
+    min-height: 0;
 }
 
 /* 2-2 */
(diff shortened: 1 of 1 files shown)


src/layouts/layouts.css
```diff
@@ -75,8 +75,8 @@
 }
 
 .layout-2-1 > div {
-    flex: 1;
-    max-height: 50%;
+    flex: 1 1 50%;
+    min-height: 0;
```
borisyankov: Is this even needed if it is 0?

nuruddeensalihu: @borisyankov  yup. Apparently the flex-basis doesn't work without it. It took me long time to realised that. You can try 😉

PR #559, opened 2016-09-21
Add intl support for browser

DanSallau:
Add support for intl polyfill.


borisyankov:
Have you tested that? Seems suspicious that it will work.
Since the request itself checks the browser agent string it will still check it.

Another issue with this logic is that if Intl is registered but others are missing, this will not run.

nuruddeensalihu:
Some browsers have a build in support for Intl(Chrome, Firefox, Yandex etc). Others don't e.g Safari. I have tested across the browsers and didn't get any issue. " if Intl is registered but others are missing, this will not run" which others do you mean @borisyankov . I think this logic should suffice. 😉 . Have tested many times.

borisyankov:
The polyfill is not just for Intl. It has about 10 other polyfills.

nuruddeensalihu:
@borisyankov  then can't we split them into individual polyfills perhaps ? The yandex browser seems to be having difficulty loading the polyfill which does not need to begin with. Also if the polyfill contain 10 others. How sure are we that the issue raised comes from Intl and not something else ? I m not sure about the other polyfills nonetheless i tested several times and this solution seems to works fine.

PR #582, opened 2016-10-07
Reduce 2/3 data when it's mobile

qingweibinary:
Do you think this make sense?


qingweibinary:
It does make mobile chart significantly faster

PR #587, opened 2016-10-17
Qingwei/purchase perf

qingweibinary:


src/trade-params/TradeParams.js
```diff
@@ -87,11 +81,11 @@ export default class TradeParams extends PureComponent {
 
     onPurchase = () => {
         const { index, proposal } = this.props;
+
+        document.getElementById(`trade-param${index}`).classList.add('greyout');
```
borisyankov: Really, direct DOM manipulation?
Do you want this to be your legacy? :D

PR #598, opened 2017-06-07
fix tnc link in upgrade account

nazaninreihani:


src/upgrade/UpgradeCard.js
```diff
@@ -137,7 +137,11 @@ export default class UpgradeCard extends PureComponent {
 			phone, termsAndConditions, progress, serverError, validatedOnce, statesList } = this.state;
 		const { residenceList } = this.props;
 
-		const firstNameIsValid = firstName.length >= 2;
+		const boot = storage.hasOwnProperty('boot') && storage.boot ? JSON.parse(storage.boot) : false;
+		const language = boot && boot.hasOwnProperty('language') && boot.language ? boot.language.toLowerCase() : 'en';
+    const linkToTermsAndConditions = `https://www.binary.com/${language}/terms-and-conditions.html`;
```
arnabk: @nazaninreihani Let me know what do you think about writing the code in this way

`
const boot = storage.boot ? JSON.parse(storage.boot) : '';
const language = boot.language ? boot.language.toLowerCase() : 'en'
`

nazaninreihani: yes this is cleaner code, I will change it , thanks

PR #600, opened 2017-06-08
add missing strings to translate files

nazaninreihani:


arnabk:
@nazaninreihani I figured out that we are not supposed to change the translation files under _constants/po/*.js

If we need new translation, we have to checkout translation branch and add all our requirements into *.po and *.pot files. Once you add those translation requirement in those files, weblate system will translate them and we have to run `sh fetch-translation.sh` command in `dev` branch to generate the new .json and .js file out of those translated files. 

Please do not worry about the changes you submitted in those .js files. They will get overidden by the command mentioned above. For now, could you please add those empty translation strings in those .po files and then submit the PR to translation branch? I will take care of all other tasks from there.

nazaninreihani:
sure I will do this

PR #601, opened 2017-06-12
Self exclusion fix

nazaninreihani:


arnabk:
@nazaninreihani 
1. This branch has conflicts, please resolve them
2. I am not sure if I am ok with this type of implementation of validation. Please let me know if you found this implementation somewhere else where copied the idea from. Otherwise we should use something better like validatejs. 

Let me know what you think.

nazaninreihani:
I agree, I'm not okay with this large objects and messy code, I try to find
another way to implement this. Is it okay to add any library in this
project? in case that it makes it cleaner and accurate.

arnabk:
@nazaninreihani I have used this library before and it is really good https://github.com/rickharrison/validate.js/

If you know anything better than this, please go ahead and use it.

arnabk:
I see conflicts in these .js files. Please revert the changes and follow my comments below - 





@nazaninreihani I figured out that we are not supposed to change the translation files under _constants/po/*.js

If we need new translation, we have to checkout translation branch and add all our requirements into *.po and *.pot files. Once you add those translation requirement in those files, weblate system will translate them and we have to run `sh fetch-translation.sh` command in `dev` branch to generate the new .json and .js file out of those translated files. 

Please do not worry about the changes you submitted in those .js files. They will get overidden by the command mentioned above. For now, could you please add those empty translation strings in those .po files and then submit the PR to translation branch? I will take care of all other tasks from there.

PR #631, opened 2017-08-10
Token of undefined issue in login

nazaninreihani:


www/boot.js
```diff
@@ -31,8 +31,8 @@
             var account = objURL['acct' + i],
                 token = objURL['token' + i],
                 currency = objURL['cur' + i];
-            if (!account || !token || !currency) break;
-            accounts.push({ account: account, token: token, currency : currency, });
+            if (!account || !token) break;
+            accounts.push({ account: account, token: token, currency : currency || '', });
```
arnabk: @nazaninreihani We might have to think about it more. I have seen this account.currency being used in multiple places. Not sure if it has some major impact allowing to proceed with empty currency. Could you please check the previous code before my changes if it was earlier allowed to pass through with currency being none? That way we would be safe to merge this change

nazaninreihani: sure

arnabk:
@nazaninreihani Are you still working on this ?

nazaninreihani:
@arnabk I mentioned you before, before the PR of adding currency to login we had SET_DEFAULT_CURRENCY so what's your idea? removing currency or adding SET_DEFAULT_CURRENCY to handle it again?

nazaninreihani:
@arnabk is it necessary to include currency in login process?

PR #640, opened 2017-08-30
Fix travis

aminroosta:


arnabk:
@aminroosta @ashkanx What was broken in travis ci? According to webpack conf, everything is copied into www and not into dist. Have we tested this change properly?

PR #641, opened 2017-09-04
add maltainvest and real account opening

nazaninreihani:


src/upgrade/UpgradeCard.options.js
```diff
@@ -0,0 +1,389 @@
+export const getOptions = () => {
+  const options = {
```
arnabk: @nazaninreihani I am not sure if we have to export a function. Can we not just do `export default { all your options data }`?

src/MultiSelect/MultiSelect.js
```diff
@@ -0,0 +1,30 @@
+import React, { PureComponent } from 'react';
+import Select from 'react-select';
```
arnabk: @nazaninreihani I am not sure, but curious. Is this supposed to be in binary-components? I do not see any other components in next-gen project.

arnabk:
@nazaninreihani This is a very big change. I do not see any tests associated with any of your changes. Are we ok without test changes?

src/MultiSelect/MultiSelect.js
nazaninreihani: No, I just wrote it for next-gen to use it in some other parts of the app too, it would be better if it was in binary-components, this is where you wanted me to implement an HOC, isn't it fine already?

nazaninreihani:
@arnabk yes you're right, I will add tests

src/MultiSelect/MultiSelect.js
arnabk: I feel this belongs to binary-components. Please submit it over there and I will get it merged.

arnabk:
@nazaninreihani Could you please list all the requirements that was implemented with this PR? This is a big change and would require some time for me to review it. I do not have the background of the functional requirement

PR #651, opened 2017-09-14
Fix ru language

aminroosta:


src/_constants/texts.js
```diff
@@ -22,6 +22,9 @@ export default ln => {
     const relatedText = t[ln];
     for (var key in relatedText) {
         if (relatedText.hasOwnProperty(key) && relatedText[key] instanceof Array) {
+            if(relatedText[key][1] == " ")
```
arnabk: @aminroosta I am not sure if this is what you wanted to do `relatedText[key][1] = relatedText[key][1].trim()`. There could be multiple white spaces?

src/_constants/po/ru.js
```diff
@@ -989,7 +989,7 @@ export default {
     ],
     "Already have an account?": [
         null,
-        " ?"
```
arnabk: @aminroosta With my fixes in, please check if this is still needed.

src/_constants/texts.js
aminroosta: @arnabk yea, closed the pr in favor of your changes ... this pr was not a solution.

PR #669, opened 2017-10-18
fix regex pattern and error message for some fields in profile and ac…

nazaninreihani:
…count openings


arnabk:
Have you run extract.py to get these into translation?

nazaninreihani:
@arnabk in translation branch?

arnabk:
@nazaninreihani 
You should add your string in this file
https://github.com/binary-com/binary-next-gen/blob/translations/build/translations/messages.pot

Note - **translation** branch in the URL.

That's all you have to do. If you think that the sentence that you removed is no longer being used, then its better to remove that sentence from message.pot file. That way we can also clean up message.pot

Once those changes are done, everything will be taken care of by the build system and the translation team.

PR #672, opened 2017-10-20
bug fix: Contract panel become partially transparent when error occurs.

bruce-binary:
When we make a purchase, whether or not it is successful or failed, it is still completed. Therefore the onPurchaseDone callback should be placed in finally block.


aminroosta (approved):
@arnabk LGTM 👍

PR #679, opened 2017-10-23
fix bug: beta version gets redirected to main app

bruce-binary:
This fix replaces BrowserHistory with HashHistory, which adds a "#" symbol to all URLs.


src/sidebar/AccountMenuItem.js
```diff
@@ -17,7 +17,8 @@ export default class AccountMenuItem extends PureComponent {
 		} else if (window.location.href.indexOf('/beta') === -1) {
```
arnabk: @brucebinary How about using `window.BinaryBoot.isBeta` ?

arnabk:
@brucebinary I was wondering, if instead of doing beta related changes all over or creating our HOC like `Link`, how about doing change in one place which is to introduce IndexRoute (in https://github.com/binary-com/binary-next-gen/blob/master/src/_routes/index.js) and map it to `/`. If we do that, in the IndexRoute class, we can check `windows.BinaryBoot.isBeta` and redirect to `/beta/` everytime someone redirects to `/`.

For example, if someone redirects to `/` and we find that isBeta is true, we redirect to `/beta`.
If someone redirects to `/#abc` and we find that isBeta is true, we redirect to `/beta#abc`

The reason I am proposing this is to find out a way to get this project to work in subfolder/namespace in case the requirement comes in future. As long as we have one place to deal with, then we are good to take it to any folders.

Not sure if my idea has some technical flaws.

I will add @aminroosta here in case he has thoughts on this too.

Notes - change from `BrowserHistory` to `HashHistory` is needed and we also have to remove all the code where there is harding code of `beta`

src/sidebar/AccountMenuItem.js
brucebinary: Ops. I'll have that changed.

brucebinary:
Sounds like a less hackish solution than what we have. If I understand correctly this requires intercepting the URL before it is processed. I'll take a look at it and get back to you.

brucebinary:
So as it turns out, we don't need to redirect to 'beta', cause if we set the app inside the beta folder (www/beta), the app will use that subfolder as the base path. This fix makes things much cleaner and simpler.

Note that because we don't have an AppId for localhost beta, binary web service will redirect localhost/beta to localhost.

aminroosta (approved):
@arnabk LGTM 👍

PR #681, opened 2017-10-24
bug fix: scroll not working in self-exclusion settings

bruce-binary:
The css rule to apply on the individual settings card is: "> div:nth-child(2)". Nesting the form under div and setting the div with the appropriate class fixes the styling issue.


aminroosta (approved):
@arnabk LGTM 👍

PR #682, opened 2017-10-25
bug fix: hidden timeout validation error message.

bruce-binary:
Steps to reproduce:
1. Go to Settings > Self Exclusion, and set "Time out until date"
2. Press "Update" and nothing happens.

The QA who initially reported this thought it was successfully updated but did not show up in the app, but in actuality nothing actually did get sent to the Binary web service. There was a validation error, but it was hidden. However, we can show this field by entering values into "Time out until time" field.

This strange behavior happens because the validation error message for a particular field only gets shown after the user enters something (in codebase it is referred as "touched") to that field.

This fix works by showing the error message when either touched.timeout_until_date or touched.timeout_until_time is being "touched". In addition, another validation is added to ensure that if user enters into "Time out until time" he is required to fill "Time out until date". Without this addition, filling out just "Time out until time" doesn't do anything but displays "Settings Updated".

### DIFF
--- a/src/settings/SettingsSelfExclusion.js
+++ b/src/settings/SettingsSelfExclusion.js
@@ -150,6 +150,7 @@ export default class SettingsSelfExclusion extends PureComponent {
       max_30day_turnover, max_30day_losses, max_open_bets, session_duration_limit,
       exclude_until, timeout_until_date, timeout_until_time, success, serverError, touched, errors } = this.state;
     // const wrongExcludeUntillTime = isValidTime(timeout_until_time);
+    const timeout_until = touched.timeout_until_date || touched.timeout_until_time;
     return (
       <div className="settings-self-exclusion">
         <form onSubmit={this.onFormSubmit}>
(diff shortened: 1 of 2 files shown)


brucebinary:
@arnabk I noticed in the codebase you left a TODO:
```
      /*
        This is not a recommended practice to have nested object in state.
        I am not sure why you need this object. Are you trying to find out
        whether a value has been changed or not? You could easily do that by
        calling !isEqual(this.props.max_balance, this.state.max_balance)
       TODO
      */
```

All of the settings have this "touched" object. It is to selectively render <ErrorMsg /> only when the user "touches" the field (types stuff into it). We could replace "touched" with that expression you suggested, but that means we should also replace it in the other files for consistency sake.

arnabk:
@brucebinary Yes, I agree. I was not fond of that code right from the beginning. I don't remember if it was already there or @nazaninreihani recently introduced it. @nazaninreihani could you confirm if we are not going to miss any scenario if we do what I left in the TODO comment?

brucebinary:
I retract my earlier statement about "All of settings". It wasn't all of them. There are discrepancies in implementation on how `<ErrorMsg />` and `<ServerErrorMsg />` is selectively rendered in different files. "touched" only appeared about 4 months back.

arnabk:
@brucebinary I have merged this PR because this is to fix something else. But if you want to improve this code, feel free to add one more task

brucebinary:
Aite.

nazaninreihani:
@arnabk right no I don't remember why I didn't change that code, so maybe I just missed, @brucebinary you can go ahead and fix it, thanks

PR #683, opened 2017-10-25
Add validation check for trading time.

bruce-binary:
I know that other components with validation code have an additional "*.validation.config.js", but just for one validation check I find it a hassle so it's all done in the same file. 

Also there is a lot of duplicate code in validation files. A lot of logic here can be shared - I could work on refactoring that as separate pull request.

### DIFF
--- a/src/trading-times/TradingTimesCard.js
+++ b/src/trading-times/TradingTimesCard.js
@@ -1,17 +1,12 @@
 import React, { PureComponent } from 'react';
-import moment from 'moment';
-import validate from 'validate.js/validate.min';
 import head from 'lodash.head';
+import validate from 'validate.js/validate.min';
 import { dateToDateString, todayLocaleString, oneYearAfterStr } from 'binary-utils';
 import { ErrorMsg } from 'binary-components';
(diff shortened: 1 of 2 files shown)


arnabk:
@brucebinary Why is it an overhead to just move the code that you have written in the same file, to a file called *.validation.config.js? I believe in JS we are used to writing single file per function too. My worry is, it starts with one method and then grows so fast that we put another refactoring effort to reduce the main file length.

brucebinary:
I see your point. I'll work on refactoring the validation code to a separate file and update this request.

brucebinary:
hm. This does meant errors is a nested object in state. But It's to keep consistent with other files.

I'm also wondering why use head() function? a[0] does the same thing. The library is only a few hundred bytes though.

brucebinary:
Actually it might be better to use head(). Just to be consistent. Thoughts?

arnabk:
@brucebinary You asked a genuine question and that has been discussed in lodash github project.
https://github.com/lodash/lodash/issues/2732

However in this case, its all about consistency.

PR #684, opened 2017-10-26
fix: upgrade account redirects to beta

bruce-binary:


aminroosta (approved):
@arnabk LGTM 👍

arnabk:
@brucebinary Does redirectUrl make sense? I wonder if this variable should be named baseUrl or rootUrl because really that is what it is. I am merging this, but something you can think about later and if agreed, can send me a separate PR

brucebinary:
Yea baseUrl sounds more appropriate. If @aminroosta is cool with that then I'll send another PR with that refactoring.

PR #685, opened 2017-10-26
fixed validation in self-exclusion

nazaninreihani:


arnabk:
@nazaninreihani Do we need anything in translation?

nazaninreihani:
@arnabk I will check translation, BTW we have many missing strings from before in translations

arnabk:
@nazaninreihani Can you point to missing translations? Will debug if there is anything wrong

PR #686, opened 2017-10-26
fix: viewport shrinks when keyboard pops up in android browsers

bruce-binary:


arnabk:
@brucebinary Do you want to add `TODO` to this comment so that we can look into this later when we get time(since you are calling it `hack`)

brucebinary:
Seems like a necessary hack to me... it's very odd code, but there's currently no other workaround for this. Let me explain: the reason why the viewport shrinks is because of this meta tag in the `<head />` of our `index.html`:

```
<meta name="viewport" content="width=device-width, height=device-height, initial-scale=1">
```

In android devices, why the keyboard pops up, the `device-width` and `device-height` changes. Because we set the viewport to use those variables, the viewport shrinks (causing the icons and text to shrink as well) when the keyboard shows up. The funny thing is `device-width` and `device-height` remain consistent with every other non-android platform. So the solution I found was just to hardcode the width and height of the device.

Or maybe I shouldn't call it a hack?

More on this: https://stackoverflow.com/questions/32963400/android-keyboard-shrinking-the-viewport-and-elements-using-unit-vh-in-css

PR #689, opened 2017-10-30
Webpack refactor

bruce-binary:
This PR introduces:
 * Prior to this PR, production and dev webpack settings are separated using `web === 'production'` conditions. This is now simplified by splitting to the config file to `webpack.common.js`, `webpack.prod.js` and `webpack.dev.js`, as suggested in [the webpack documentation](https://webpack.js.org/guides/production/).
 * `.babelrc` is refactored to a separate file instead of being fused in `package.json`.
 * Remove moment.js locales - 2.9Mb -> 2.73Mb (170Kb reduction in size). 
 * updated svg-react-loader.

In dev mode, the app.js is huge (15Mb), but for good reason: I've set `devtool` attribute in `webpack.dev.js` to use `eval-source-map`, which allows accurate line number mapping when using web browser debugger during development.


arnabk:
@brucebinary Please resolve the conflict

brucebinary:
@arnabk aite just resolved it.

brucebinary:
ok. I have no idea why there was conflict after I already resolved it, but I pushed a commit where I just remove a space and now there's no conflicts. 

I think next time I will only resolve conflicts locally.

PR #691, opened 2017-10-31
fix translation for self-exclusion

nazaninreihani:


arnabk:
@nazaninreihani Please keep an eye on message.pot and if you think that some texts are getting removed automatically, we have to find out which automatic build process is doing it. If you notice the issue again, let me know.

PR #692, opened 2017-10-31
Refactor validation code

bruce-binary:
Here are my proposed changes to how the validation code could be simplified. A lot of refactoring is required, so I only started with `SettingsSelfExclusion.js` as a preview of what to expect (I also made sure the unit tests passed so no worries) if this is applied to the rest of the project (only parts that used `validate.js` will be affected by this refactor.

## validateExtend.js
Consolidate all custom validators of [`validate.js`](https://validatejs.org/) into `validateExtend.js`. This file must be called prior to any validation being called, and is placed in `index.js` and added to jest's setup files.

## ValidationManager
Notice that the `touched` dictionary is removed from the form. The responsibility of keeping track of interacted fields is now passed to a new class `ValidationManager`. Now we call `validateFieldAndGetNewState` method to get the updated form data and its corresponding errors. Since this method selectively chooses which fields to validate based on what is being interacted, a separate `touched` dictionary is no longer required. To validate all fields (say when we submit the form), we can use `validateAll` method. 

## Misc Bug Fixes
This PR also fix some bugs with SettingsSelfExclusion:
 * When timeout until date and time is set, settings get updated but page does not refresh. Expected behavior is that page should refresh and user will be shown dialog box "you have been excluded until...".
 * When user updates a setting and go to another tab, "setState cannot be called on unmounted components" warning appears in console after 3 seconds.
 * fix moment.js deprecation warnings.


arnabk:
@brucebinary This is much better. I will leave this PR open for 1-2 days in case @aminroosta and @nazaninreihani want to comment

nazaninreihani (approved):

aminroosta:
@arnabk LGTM 👍

aminroosta (approved):

arnabk:
@brucebinary Please resolve conflicts

brucebinary:
Aite. So I take that everyone is in agreement with these changes. I will work on refactoring the remaining affected files and update this pull request. Should be done latest tomorrow.

arnabk:
@brucebinary It shows conflict again.

brucebinary:
@arnabk I'm currently working on fixing a bug (trading on closed markets). Will finish the refactoring once I'm done.

brucebinary:
@arnabk I've finished refactoring. It's a big change, but thankfully there's unit tests right? (;

In the recent commits, I also added validation for all affected fields when user clicks submit. Currently what happens is that it seemed to fail silently, but there's actually an error message on the top. I've added a trello card for QA to verify this change.

brucebinary:
@arnabk also this branch is branched-off from async-webtrader-charts since merging them was a pain in the ass. [This pull request](https://github.com/binary-com/binary-next-gen/pull/695) needs to be merged first before this.

src/upgrade/UpgradeToMaltainvestCard.js
```diff
@@ -46,119 +47,66 @@ export default class UpgradeToMaltainvestCard extends PureComponent {
...
+    const val = e.target.checked ? 1 : 0;
+    const s = this.validationMan.validateAndGetNewState('accept_risk', val, this.state.formData);
+		this.setState({ ...s, hasError: false });
   }
 
   onPEPDeclarationChanged = (e: SyntheticEvent) => {
-    const PEPDeclaration = e.target.checked ? '1' : '0';
-    this.setState({ PEPDeclaration, touched: { ...this.state.touched, [e.target.id]: true }, });
+    const val = e.target.checked ? 1 : 0;
+    const s = this.validationMan.validateAndGetNewState('PEPDeclaration', val, this.state.formData);
```
nazaninreihani: @bruce I think here PEPDeclaration is in formData which is going to be in request, right?
 but it's just a checkbox which should be checked to allow the users to upgrade their account, it's not included in the request, so when we include it in request user will get an error in response that it's not defined in the schema.
could you please check this again?

brucebinary: I'm currently getting access to a QA box to test this, but in the mean time this is a simple fix. I'll put up a PR for this by today.

brucebinary: This should be fixed in https://github.com/binary-com/binary-next-gen/pull/716

PR #695, opened 2017-11-01
Load webtrader charts asynchronously

bruce-binary:
So following @aminroosta instructions I have the webtrader-charts and highstock library separated from the `app.js` (prior to this `app.js` is 2.75Mb):
```
                         Asset      Size  Chunks                    Chunk Names
    webtrader-charts.bundle.js    906 kB       0  [emitted]  [big]  webtrader-charts
                        app.js   1.75 MB       1  [emitted]  [big]  app
```
Here is a before and after of the network performance tool:
**Before**

**After** (check it out at https://brucebinary.github.io/binary-next-gen)

Minor stuff also in this PR:
 * to use the `async` with `await import` syntax I added babel preset `stage-3`. 
 * yargs is only used in `/build` folder, so it is moved there.


src/contract-details/ContractChart.js
```diff
@@ -67,35 +66,43 @@ export default class ContractChart extends PureComponent {
...
-      instrumentName: symbolName,
-      showInstrumentName: true, // default is false
-      showOverlays: false, // default is true
-      showShare: false, // default is true
-      start: (purchase_time || date_start) - margin,
-      end: sell_time && +sell_time + margin || (exit_tick_time ? exit_tick_time + margin : 'latest'),
-      hideCurrentPrice: true,
-      // timezoneOffset: 0,
-    });
+    (async () => {
```
aminroosta: @brucebinary use async lifecycle methods instead of an iife, see [this blog post](https://medium.com/front-end-hacking/async-await-with-react-lifecycle-methods-802e7760d802)

src/trade/trade-chart/TradeViewChart.js
```diff
@@ -60,15 +54,27 @@ export default class TradeViewChart extends PureComponent {
...
+
   destroyChart() {
     this.chart && this.chart.actions.destroy();
     this.chart = null;
   }
 
   componentDidMount() {
-    const trade = this.props.tradeForChart.toJS();
-    this.initChart(trade);
+    (async () => {
```
aminroosta: same here

aminroosta:
@brucebinary `285KB (webtrader-charts) +459KB(application code) = 744KB` which is more than when we used a single bundle `690KB(app.js)`

src/trade/trade-chart/TradeViewChart.js
brucebinary: huh. Great suggestion @aminroosta . Definitely simplifies the code a bit.

brucebinary:
@aminroosta on the increase in total size of the js files. IMHO it is a worthy trade off, since the application loads up first before the webcharts. It's a 1 second difference as above, but the impact could be bigger with slower connections.

arnabk:
@brucebinary Please resolve conflicts

aminroosta:
@brucebinary Yes it should be fine, i think that's probably because of the way gzip works.
It's algorithm works better with larger files.

arnabk:
@brucebinary @aminroosta Let me know if this is ready to me merged

brucebinary:
@arnabk it's ready

aminroosta (approved):
@arnabk LGTM 👍

PR #696, opened 2017-11-01
React 16 upgrade

bruce-binary:
upgraded to react v16.

TDLR: This PR shrinks `app.js` by 80Kb.

This upgrade shaves off 30Kb from `app.js` (from 1.75Mb to 1.72Mb; this is with async loading of webtrader-charts). React v16 is supposedly faster, but I didn't notice any difference. 

Also if we agree to upgrade, [binary-components](https://github.com/binary-com/binary-next-gen) should be upgraded to react 16 as well.

Some porting notes:
* if `shouldComponentUpdate` is used in `PureComponent` the class needs to be changed to `Component`.
* react-redux is upgraded to v5 to be compatible with react 16
* as of react 16, `react-addons-perf` is no longer supported and thereby removed in this commit (see https://reactjs.org/docs/perf.html)

Smaller addons:
 * added `ModuleConcatenationPlugin`. With async loading of webtrader-charts and react 16, this shrinks `app.js` from 1.72Mb to 1.67Mb.
 * Shifted `ExtractTextPlugin` to be only used in production. In dev, we want HMR with stylesheets so we can immediately see changes when we update our `*.scss` files.


arnabk:
@brucebinary I would recommend upgrading binary-components too, if possible

brucebinary:
@arnabk aite. I'll work on porting that.

PR #705, opened 2017-11-13
 Trade params datetime + stake amount fields now retains its values...

bruce-binary:
...*unless* there are input errors, in which next-gen will revert to a default date. Switching back and forth "Later" and "Now", or choosing different trade types, or refreshing the page should also retain its values. This should also work for multiple trades. 

For something so simple it is a little unusual to have so much changed. So some explanation is warranted. The existing code was confusing to me so I may understand it incorrectly. Feel free to critique as required:

## Forced Render

The code has a `UPDATE_TRADE_UI_STATE` action which triggers a force render. It is used in only 2 places: the duration picker and the forward time picker. Though I was never able to find the implementation that does this, I observe the affects: this "force rendering" destroys and recreates the `<forwardStartingOptions >` component each time we switch back and forth "Later" and "Now". Same applies for the duration picker. This is why the values of that component is never retained even though it is being stored in local storage; the component is constantly being destroyed and recreated. Though explicitly programmed by the previous programmer, I find it very odd that this is required at all. It’s definitely not standard practice in React code.

Why force render? 

Well without it, the datetime field values refuse to change no matter what we set. Looking at the code, I noticed this was because `defaultValue` was used to set the value instead of `value`. Of course you will never change the actual value - you only change the default value! So in order for the datetime field to match with what gets passed to the server, the value is reset to the default value, by destroying and recreating the component.

This is an unnecessarily complicated solution to a simple problem.

I think it is more sensible to keep convert the fields to a [controlled component](https://reactjs.org/docs/forms.html#controlled-components), and use datetime values from the state ("the single source of truth").

## GMT Offset Bug

Once the force rendering is removed, a funny GMT offset bug surfaces. The datetime shown in the trade params and the datetime actually used to make the purchase is different. This is corrected when the user changes the datetime fields, and comes back when when the user switches between "Now" and "Later".

This is because the GMT offset is calculated (in a cryptic fashion) in `handleStartDateChange` and `handleStartTimeChange` function but not in `handleStartEpochChange` function. But why do we need 3 functions that do practically the same thing? 

So I removed `handleStartDateChange` and `handleStartTimeChange` and just use `handleStartEpochChange`. I then also decided (and we can debate on this) that the GMT offset calculation should in done in *when* we send the proposal. We can then assert that if we pass our date from our own timezone to the subscribe proposal function, it will convert to GMT time before sending to the server. 

To be clear, the offset bug was not caused by removing forced render; it was always there, but never surfaces because the component was always destroyed and recreated, resetting it to the default value.


arnabk:
@brucebinary I think these are sensitive changes. I think the best way to make sure that nothing related to trade is broken is to use https://github.com/binary-com/binary-next-gen-nightwatch and find out if all test for trade modules are passing. If something is missing, then we can add there too.

Without getting deep into your changes and basing just off of what you have explained, the only reason I could think of why we should be removing the component is to make sure that if something two different components are bound to the same variable and in reality only is supposed to be rendered at a time, this situation could lead to problem. And so I am not sure if I would agree to `It’s definitely not standard practice in React code.`. It should be case by case basis.

The other question I have is, why are we dealing with TZ(Meaning offset)? Can we not just stick to GMT? Meaning we display GMT and we send GMT to server. That way, everything is straight forward.

Note - My comments are on very high level without getting deep down into the code changes.

brucebinary:
@arnabk I did not know https://github.com/binary-com/binary-next-gen-nightwatch existed until today... I'll look it up and get back to you. 

I'm not sure what you mean by:

> the only reason I could think of why we should be removing the component is to make sure that if something two different components are bound to the same variable and in reality only is supposed to be rendered at a time, this situation could lead to problem.

If they are bound by the same variable, then a parent component should take responsibility of that variable (as a state attribute, in which changing it results in a rerender) and control which of the 2 components gets rendered.

As for GMT, we have to deal with the offset because the server expects the time in epoch (number of seconds that have elapsed since January 1, 1970 GMT). That time varies depending on where you are; for example in Malaysia we need to add `8 * 60 * 60` seconds. This is why in production we have this code block (`DurationSaga.js`):

```
const newDayEpoch = dateToEpoch(new Date(date));
// ...
const secondsPerDay = 60 * 60 * 24;
const intraDayEpoch = dateStart % secondsPerDay;
const newDateStart = newDayEpoch + intraDayEpoch;
```
Here `dateToEpoch()` returns the `date` in unix time at 12AM with the GMT offset, and `intraDayEpoch` isolates the time from `dateStart` and adds it to `newDayEpoch`. The same can be achieved with

```
const timezoneOffsetSeconds = new Date().getTimezoneOffset() * 60;
dateStart -= timezoneOffsetSeconds;
```

However, when we save that date in local storage, we don't want to save the date with the GMT offset...

...you know what it's pretty confusing. We should be passing the date object around, and before sending to the server convert to epoch time:

```
const epoch = date.getTime() / 1000; // convert milliseconds to seconds
```

brucebinary:
@arnabk After some work I manage to get the nightwatch project running on my localhost project (the instructions are dated; accessing folders which no longer exists). It seems like it's running on an older version of next-gen, cause most of the tests failed looking for `#txtEmail` field (which is in the OAuth page). 

It was also looking for a "videos" page which no longer exists anymore.

It will take some time to get it to work. I'm opening up a trello card for this.

brucebinary:
@arnabk hm. I'm starting to agree that the changes are too sensitive. Perhaps an alternative: Retain `UPDATE_UI_STATE` and its force rendering, and only remove its function call from the datetime field. 

We can't to keep the `UPDATE_UI_STATE` function call in datetime field and still retain its values. So this would be a good compromise... whatever `UPDATE_UI_STATE` was initially designed to do...

arnabk:
@brucebinary In Webtrader, we had lot of issues dealing with trade module. It's not because we did something wrong, it's because the trade module API is not straight forward. Trade module is sensitive because we deal with client's money and if we want to not get hit by those issues again, I would recommend it is worth the effort to fix nightwatch project and add more int-tests for the trade module before releasing this piece of code(or any trade module related code)

brucebinary:
@arnabk aite. I'll have that nightwatch project fixed first before returning to this. This PR for now will be put on hold.

arnabk:
@brucebinary Since nightwatch for this project is dead, do you want me to release this? @aminroosta Since you are more experienced in trade module, could you please review this change?

brucebinary:
@arnabk @aminroosta  owh I've forgotten about this. A lot has changed in the code base since this commit... Let me get back to you guys on this.

src/duration-picker/DurationCard.js
```diff
@@ -66,7 +66,7 @@ export default class DurationCard extends PureComponent {
                     <NumericInput
                         className="numeric-input"
                         integer
-                        defaultValue={duration}
+                        value={duration}
```
brucebinary: requires https://github.com/binary-com/binary-components/pull/45

arnabk:
@ashkanx @aminroosta Please review and release this. This PR is needed for https://github.com/binary-com/binary-next-gen/pull/705

brucebinary:
@aminroosta good to merge? 👍

aminroosta:
@arnabk LGTM 👍

brucebinary:
@arnabk I've fixed the merge conflicts

PR #709, opened 2017-11-14
Odd autocomplete implementation with NumericInput

bruce-binary:
So a QA reported this funny autocomplete going completely out of position:

A curious thing is that it only affects android devices. In desktop and in iOS devices this autocomplete doesn't appear. What is this autocomplete doing in the first place?

Turns out it's actually from a `<datalist>` from the stake input. Unfortunately the list attribute in `NumericInput` is hardcoded to be "values" in binary-components, so if we have multiple value lists there's going to be ambiguity; ideally we want the id for each datalist to be unique. So passing a `valueList` prop to `NumericInput` is a bug waiting to happen.

But I have a simpler solution. Why do we need hardcoded payout values in the first place? Binary-static doesn't use it.

### DIFF
--- a/src/payout-picker/StakeCard.js
+++ b/src/payout-picker/StakeCard.js
@@ -3,7 +3,6 @@ import debounce from 'lodash.debounce';
 import { NumericInput, Label } from 'binary-components';
 import { actions } from '../_store';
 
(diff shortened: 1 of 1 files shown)


arnabk:
@brucebinary Your tests passing for this change or there were no unit tests for this?

brucebinary:
I only removed autocomplete. I don't think there are unit tests for that.

PR #710, opened 2017-11-15
Use frontend.binaryws as websocket endpoint

aminroosta:


arnabk:
@nazaninreihani Why do you have to do so much change for just URL change (https://github.com/binary-com/binary-next-gen/pull/707) ? Please look into this PR that @aminroosta submitted. Are we missing anything here?

PR #714, opened 2017-11-18
Add compatibility with endpoint chrome extension

nazaninreihani:


www/boot.js
```diff
@@ -14,6 +14,34 @@
...
+        window.BinaryBoot.appId = 3604;
+      } else if (window.BinaryBoot.isBeta) {
+        window.BinaryBoot.appId = 4343; //This is for BETA release
+      }
+      localStorage.setItem('config.default_app_id', window.BinaryBoot.appId);
+      return window.BinaryBoot.appId;
+    };
+
+  const getAppId = () =>
+    window.localStorage.getItem('config.app_id') || getDefaultAppId() || 1001;
```
brucebinary: I'm guessing storing the app id in local storage is an optimization instead of needing to figure out the app id each time the page refreshes. Why is the app id for production defined in this line? It could be part of `getDefaultAppId()` right?

Actually I took notice cause it broke the automatic personal github pages deployment script. It looks for this line:

```
window.BinaryBoot.appId = 1001; //This is for PROD release
```

I could update the script to detect from the line above, but you know... `getDefaultAppId() || 1001` is pretty cryptic. It is not clear that 1001 is the app id of the production release.

PR #716, opened 2017-11-21
Validation refactor fix

bruce-binary:
Some bugs were introduced by the refactor:

- https://github.com/binary-com/binary-next-gen/pull/692#pullrequestreview-77616965
- https://github.com/binary-com/binary-next-gen/commit/c30130949a124a90da9cf27492509ca7bd05547a#commitcomment-25707799

This PR fixes them. Kudos to @nazaninreihani for bringing them up.

However, I feel this is something that could be covered by tests. I'm not sure if react enzyme can do this or do I need to use selenium. @nazaninreihani what do you think?

### DIFF
--- a/src/settings/SettingsSelfExclusion.js
+++ b/src/settings/SettingsSelfExclusion.js
@@ -24,21 +24,9 @@ export default class SettingsSelfExclusion extends PureComponent {
 
   constructor(props) {
     super(props);
+    const formData = this.getFormData(props);
     this.state = {
-      formData: {
-        max_balance: props.max_balance,
-        max_turnover: props.max_turnover,
-        max_losses: props.max_losses,
-        max_7day_turnover: props.max_7day_turnover,
-        max_7day_losses: props.max_7day_losses,
-        max_30day_turnover: props.max_30day_turnover,
-        max_30day_losses: props.max_30day_losses,
-        max_open_bets: props.max_open_bets,
-        session_duration_limit: props.session_duration_limit,
-        timeout_until_time: props.timeout_until,
-        timeout_until_date: props.timeout_until,
-        exclude_until: props.exclude_until,
-      },
+      formData,
       errors: {},
       serverError: false,
       success: false,
@@ -47,6 +35,28 @@ export default class SettingsSelfExclusion extends PureComponent {
     this.constraints = getConstraints(this.state.formData);
     this.validationMan = new ValidationManager(this.constraints);
   }
+
+  getFormData(props) {
+    return {
+      max_balance: props.max_balance,
+      max_turnover: props.max_turnover,
+      max_losses: props.max_losses,
+      max_7day_turnover: props.max_7day_turnover,
+      max_7day_losses: props.max_7day_losses,
+      max_30day_turnover: props.max_30day_turnover,
+      max_30day_losses: props.max_30day_losses,
+      max_open_bets: props.max_open_bets,
+      session_duration_limit: props.session_duration_limit,
+      timeout_until_time: props.timeout_until,
+      timeout_until_date: props.timeout_until,
+      exclude_until: props.exclude_until,
+    };
+  }
+
+  componentWillReceiveProps(nextProps) {
+		const formData = this.getFormData(nextProps);
+		this.setState({ formData });
+	}
 
   onEntryChange = (e: SyntheticEvent) => {
     const s = this.validationMan.validateFieldAndGetNewState(e, this.state.formData);
--- a/src/settings/SettingsUserInformation.js
+++ b/src/settings/SettingsUserInformation.js
@@ -26,25 +26,34 @@ export default class SettingsUserInformation extends PureComponent {
 
 	constructor(props) {
 		super(props);
-
+		const formData = this.getFormData(props);
 		this.state = {
-			formData: {
-				address_line_1: props.address_line_1,
-				address_line_2: props.address_line_2,
-				address_city: props.address_city,
-				address_state: props.address_state,
-				address_postcode: props.address_postcode,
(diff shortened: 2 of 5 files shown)


arnabk:
@nazaninreihani Do you want to give it :+1: for merge?

nazaninreihani:
@brucebinary I thinke2e testing is a good idea and selenium is the best option, but let's know @aminroosta and @arnabk idea

src/settings/SettingsSelfExclusion.js
```diff
@@ -48,6 +36,28 @@ export default class SettingsSelfExclusion extends PureComponent {
...
+      max_30day_losses: props.max_30day_losses,
+      max_open_bets: props.max_open_bets,
+      session_duration_limit: props.session_duration_limit,
+      timeout_until_time: props.timeout_until,
+      timeout_until_date: props.timeout_until,
+      exclude_until: props.exclude_until,
+    };
+  }
+
+  componentWillReceiveProps(nextProps) {
```
nazaninreihani: @brucebinary I think componentWillRecieveProps is fired even when props are not changed so when using that for preventing extra rerendering we need to check if current props are changed, what do you think about checking (next props !== this.props) before doing anything?

arnabk:
@nazaninreihani We have https://github.com/binary-com/binary-next-gen-nightwatch
We use nightwatch.

src/settings/SettingsSelfExclusion.js
arnabk: @nazaninreihani "I think componentWillRecieveProps is fired even when props are not changed" are you sure about that? This is to challenge Facebook's architecture.

nazaninreihani: @arnabk it's on their own document 
https://reactjs.org/docs/react-component.html#componentwillreceiveprops

brucebinary: it's odd to need that condition. By adding that condition we are comparing the addresses of those 2 objects, which will always be different (I've check this just to be sure). Unless what you mean is to compare each individual attribute of the form data (like how it was initially implemented by @arnabk ):

```
  componentWillReceiveProps = (newProps) => {
    const newState = {};
    if (newProps.max_balance !== this.state.max_balance) newState.max_balance = newProps.max_balance;
    if (newProps.max_turnover !== this.state.max_turnover) newState.max_turnover = newProps.max_turnover;
    if (newProps.max_losses !== this.state.max_losses) newState.max_losses = newProps.max_losses;
    if (newProps.max_7day_turnover !== this.state.max_7day_turnover) newState.max_7day_turnover = newProps.max_7day_turnover;
    if (newProps.max_7day_losses !== this.state.max_7day_losses) newState.max_7day_losses = newProps.max_7day_losses;
    if (newProps.max_30day_turnover !== this.state.max_30day_turnover) newState.max_30day_turnover = newProps.max_30day_turnover;
    if (newProps.max_open_bets !== this.state.max_open_bets) newState.max_open_bets = newProps.max_open_bets;
    if (newProps.session_duration_limit !== this.state.session_duration_limit) newState.session_duration_limit = newProps.session_duration_limit;
    if (newProps.timeout_until_time !== this.state.timeout_until_time) newState.timeout_until_time = newProps.timeout_until_time;
    if (newProps.timeout_until_date !== this.state.timeout_until_date) newState.timeout_until_date = newProps.timeout_until_date;
    if (newProps.exclude_until !== this.state.exclude_until) newState.exclude_until = newProps.exclude_until;
    if (Object.keys(newState).length > 0) this.setState(newState);
  }

``` 

that would be pretty verbose, and I'm not too concern with rerenders in virtual DOM as it is pretty cheap to do. Rerenders also do not happen immediately; react may delay them and update a few changes in a single pass.

However I am willing to place the condition block back if everyone agrees this is more performant. @aminroosta your thoughts?

brucebinary: ops. I didn't know there were comments before mine. Github doesn't update this page in the background

brucebinary:
Right forgot to mention about that nightwatch repo (it uses selenium). I'll be bringing that project back up once we push a release.

src/settings/SettingsSelfExclusion.js
nazaninreihani: @brucebinary @arnabk what happens to state if componentWillReceiveProps is called unexpectedly in the middle of editing the field by the user? isn't the state going to be set to new Props and all user input missing?

arnabk: @nazaninreihani Isn't PureComponent supposed to skip the rerendering if the props are same? I thought that is the reason they introduced it instead of we doing the shallow compare.

brucebinary: @nazaninreihani I've recently given some thought to that as well. I've set a breakpoint on `componentWillReceiveProps` before, and noticed that the form data attributes were `undefined` for 3 consecutive times, and then contain data in the next 3 consecutive calls - this happens so fast that the user never notices, but it could be probable that in slow connections user can have his inputs overwritten (not with blank data but saved data).

I find this behaviour very unusual. Shouldn't the form data in props be there during initialisation and not only receive the data in the background milliseconds later? Filling the form data using `componentWillReceiveProps` is a little odd to me.

nazaninreihani: @arnabk @brucebinary Found the issue, the issue seems when we set initial state from props, after recieving props the new props are not passed to the component, so we can remove the initial value for formData from constructor and set inital values in render

nazaninreihani: @brucebinary sorry I didn't see your last commits

PR #719, opened 2017-11-21
convert es6 to es5 in boot.js

nazaninreihani:


www/boot.js
```diff
@@ -106,13 +109,12 @@
...
-      try {
+    try {
         // var config = JSON.parse(testConfig) || { };
-        const serverURL = localStorage.getItem('config.server_url');
         window.BinaryBoot.appId = getAppId();
-        window.BinaryBoot.apiUrl = serverURL ? `wss://${serverURL}/websockets/v3` : getSocketURL();
+        window.BinaryBoot.apiUrl = getSocketURL();
         window.BinaryBoot.oAuthUrl = config.oAuthUrl || window.BinaryBoot.oAuthUrl;
-      } catch (e) { }
+    } catch (e) { }
```
brucebinary: @nazaninreihani is there a reason for this empty try-catch block? I think it was here for some time before this commit. I read that this is generally bad practice; errors messages will just get swallowed up.

brucebinary: Ah. I see why it was there. Without the empty try-catch block this error comes up:
```
Uncaught ReferenceError: config is not defined
```

So pull request https://github.com/binary-com/binary-next-gen/pull/720 should fix this.

Still. Without that empty try-catch block we would have caught that error sooner.

nazaninreihani: @brucebinary yes the error is fixed in that PR, but still we may have some silent errors

PR #728, opened 2017-11-24
[mobile] disable navigator and used a styled scrollbar

bruce-binary:
Requires https://github.com/binary-com/webtrader-charts/pull/94 to be deployed.


styles/_mobile.scss
```diff
@@ -29,6 +29,34 @@
...
+    .highcharts-scrollbar-track {
+        height: $scroll-bar-height;
+        rx: $scroll-bar-height;
+        ry: $scroll-bar-height;
+    }
+
+    .highcharts-scrollbar-rifles {
+        stroke: transparent;
+    }
+
```
aminroosta: @brucebinary Could we move the styles to webtrader-charts instead?

brucebinary: I have some considerations for that... but the new scrollbar doesn't look good with the navigator:

binary_com

Unless webtrader detects mobile view and selectively chooses the style.

aminroosta: > Unless webtrader detects mobile view and selectively chooses the style

@brucebinary We can pass "isMobile: true" from next-gen to webtrader-charts.

brucebinary: @aminroosta that could work. So in that case I suggest in webtrader-charts add a variable `enableMobileView`. We could also remove the `enableNavigator` setting since enabling mobile view disables the navigator.

PR #729, opened 2017-11-27
add global function to modify the endpoint from console

bruce-binary:


README.md
```diff
@@ -25,6 +25,20 @@ yarn test
 
 Once your changes have been merged to `dev`, it will immediately deployed to app.binary.com/beta.
 
+## Change Endpoint
```
ashkanx: @brucebinary does QA really need this? We're developing a Chrome Extension for QA to change the endpoint.

brucebinary: @ashkanx QA requested for an alternative cause they can't use the chrome extension in mobile devices.

PR #733, opened 2017-11-28
Multi account support and set currency

nazaninreihani:


arnabk:
@nazaninreihani Its a lot of change. No tests class?

PR #735, opened 2017-11-30
Automatically Deploy Apk to Github Releases

bruce-binary:
I've gotten the travis script to work on my github repo, so you can see how it would look like if the script runs successfully in binary-com/binary-next-gen: https://github.com/brucebinary/binary-next-gen/releases 

It's a pretty big change on travis script - changing from `node_js` project to `android` project (granted we only support android now), so review it with the utmost scrutiny. Documentation is also added on how to modify the travis script to work with personal github pages.

The website will get deployed long before the beta version of the app does. This is because the app takes about 10 minutes to build (the website takes about 2 minutes), so I decided to have the site deployed first. I could have them deployed at the same time if you like (it would make the travis script simpler).


package.json
```diff
@@ -6,15 +6,15 @@
   },
   "scripts": {
     "test": "jest",
-    "test:full": "npm run build && npm run test:coverage",
+    "test:full": "npm run test:coverage",
```
brucebinary: Notice that I removed the build command here. Performing tests do not require the project to build IIRC. This can save us about 2 minutes from building the project. The script will eventually `yarn build` in the "build" folder, which creates a "dist" folder in the project directory. The contents of this folder is what we eventually deploy.

package.json
```diff
@@ -6,15 +6,15 @@
...
     "test:flow": "flow check --all --show-all-errors",
     "test:unit": "jest",
     "test:coverage": "jest --coverage",
     "test:watch": "npm test -- --watch",
     "test:unit-watch": "jest --watch",
     "test:coveralls": "cat ./coverage/lcov.info | ./node_modules/coveralls/bin/coveralls.js && rm -rf ./coverage",
     "start": "webpack-dev-server --open --config webpack.dev.js",
-    "clean": "rimraf dist",
+    "clean": "rm -rf dist",
```
arnabk: @brucebinary This is not going to work for Windows users

.travis.yml
```diff
@@ -14,20 +21,26 @@ branches:
   - master
   - dev
 before_install:
-  - yarn
+  # cordova currently only works with node 6.x
+  - curl -sL https://deb.nodesource.com/setup_6.x | sudo -E bash -
+  - sudo apt-get install -y nodejs
```
arnabk: @brucebinary If you use `dist : trusty`, you don't need `sudo` (if I remember correctly)

arnabk:
@brucebinary The overall change looks straight forward to me. Is there anything that you would like me to focus while reviewing?

@aminroosta Do you want to give it a quick review because this would change how the TravisEnv is build for this project.

package.json
brucebinary: owh. I forgot to mention why I changed that. The rimraf thing doesn't work in travis once I changed to the android env - not sure why. Let me take another look at that

Are we suppose to support windows?

.travis.yml
brucebinary: I remember that sudo not being needed for one of the distributions as well. I'll see if I can remove sudo to keep things simple abit

package.json
arnabk: This is a open source project. Sometimes 3rd parties fork this project and do their own thing. I also remember, one of the telecommuters was also using Windows for development. Its an on/off thing. I think @nazaninreihani might also be using Windows(not sure). So its better to support windows when possible. That `rimraf` probably does not work in windows either.

nazaninreihani: @arnabk @brucebinary I'm using ubuntu, no issue for ubuntu I think, yes?

brucebinary: travis runs on ubuntu, so I don't think so

.travis.yml
brucebinary: @arnabk nope doesn't work.

package.json
```diff
@@ -14,7 +14,7 @@
     "test:unit-watch": "jest --watch",
     "test:coveralls": "cat ./coverage/lcov.info | ./node_modules/coveralls/bin/coveralls.js && rm -rf ./coverage",
     "start": "webpack-dev-server --open --config webpack.dev.js",
-    "clean": "rm -rf dist",
+    "clean": "rimraf dist",
```
brucebinary: @arnabk so rimraf is back! Turns out in this script travis doesn't use it anymore.

PR #740, opened 2017-12-05
bug fix: unable to sell at market after refresh + portfolio does not …

bruce-binary:
…remove expired contracts

contract_id is expected to be a string, but number is returned. Alternately you could convert to number (the API is supposedly returning Number, but perl somehow converts it to string), but changing to string requires less modifications.

### DIFF
--- a/src/_actions/PortfolioActions.js
+++ b/src/_actions/PortfolioActions.js
@@ -14,8 +14,10 @@ export const subscribeToOpenContract = contractId =>
         if (!boughtContracts.get(contractId)) {
             return api.subscribeToOpenContract(contractId)
                 .then(response => {
-                    if (Object.keys(response.proposal_open_contract).length > 0) {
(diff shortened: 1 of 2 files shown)


src/_actions/PortfolioActions.js
```diff
@@ -14,8 +14,10 @@ export const subscribeToOpenContract = contractId =>
         if (!boughtContracts.get(contractId)) {
             return api.subscribeToOpenContract(contractId)
                 .then(response => {
-                    if (Object.keys(response.proposal_open_contract).length > 0) {
-                        return response.proposal_open_contract;
+                    const c = response.proposal_open_contract;
+                    if (Object.keys(c).length > 0) {
+                        c.contract_id = String(c.contract_id);
```
arnabk: @brucebinary 👏  for using String()

PR #741, opened 2017-12-05
bug fix: in iOS safari, the bottom mobile bar obstructs content.

bruce-binary:
This commit simply introduces some space of the bottom bar to take up.

### DIFF
--- a/src/_utils/Device.js
+++ b/src/_utils/Device.js
@@ -1,0 +1,17 @@
+export default class Device {
+    static isAndroid() {
+        return /(android)/i.test(navigator.userAgent);
+    }
+
+    static isAndroidApp() {
+        return window.cordova && device.platform === 'Android';
+    }
+
+    static isIOSSafari() {
+        const ua = window.navigator.userAgent;
+        const iOS = !!ua.match(/iPad/i) || !!ua.match(/iPhone/i);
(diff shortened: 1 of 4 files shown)


arnabk:
Bit confused about this change. I think is another piece of code where you are checking for iOS device. Earlier there was some code to check for Android. First, we are not supporting iOS. We are supporting OSX. Second, Can we think of putting it in a common util file so that it could be reused elsewhere, if needed. Otherwise, I am pretty sure this will end up being duplicated multiple times when different developers are working on it.

brucebinary:
@arnabk True - we are not supporting the iOS app anymore, but the web app should still run in iOS safari right?

I'll see what I can do about that refactoring out a util file.

src/_utils/Device.js
```diff
@@ -0,0 +1,17 @@
...
+    }
+
+    static isIOSSafari() {
+        const ua = window.navigator.userAgent;
+        const iOS = !!ua.match(/iPad/i) || !!ua.match(/iPhone/i);
+        const webkit = !!ua.match(/WebKit/i);
+        const iOSSafari = iOS && webkit && !ua.match(/CriOS/i);
+        return iOSSafari;
+    }
+}
```
brucebinary: @arnabk I've refactored out a separate file for device detection. Lemme know if it's good to go

PR #752, opened 2017-12-08
[cordova] bug fix: black screen appears when user upgrades virtual ac…

bruce-binary:
…count.

Best to avoid `window.location = "..."`; where baseUrl is `/`, this causes the path of the cordova app to change from `file:///android_asset/www/index.html` to `/`.


src/_data/LiveData.js
```diff
@@ -259,7 +259,6 @@ export const setAccountCurrency = (currency, store) => {
       if (r.set_account_currency === 1) {
         const loginid = state.account.toJS().loginid;
         addCurrencyToAccount(currency, loginid);
-       window.location = window.BinaryBoot.baseUrl;
```
brucebinary: @nazaninreihani I've shifted the refresh page code to `SetCurrency.js` instead, since I can't access the router from there.

src/sidebar/AccountMenuItem.js
```diff
@@ -9,14 +9,15 @@ export default class AccountMenuItem extends PureComponent {
...
+
 	switchToAccount = () => {
 		const { token } = this.props;
 		storage.setItem('account', JSON.stringify({ token }));
-		if (window.cordova) {
-			window.location.reload(true);
-		} else {
-			window.location.href = window.BinaryBoot.baseUrl;
-		}
+		this.context.router.push('/');
```
brucebinary: this change meant switching accounts will set the app to go the home page, instead of refreshing the page. I think this should be the expected behaviour when switching accounts.

arnabk: @brucebinary if the baseUrl is /beta instead of /, will it set to /beta?

brucebinary: @arnabk paths in `this.context.router` happens in the context of the app; so if the app exists in a folder `/beta`, it's root path is still `/`. This is different from using `window.location.href`, which changes the URL.

src/_data/LiveData.js
nazaninreihani: @brucebinary users can choose currency in upgrade to maltainvest process and after upgrading success a request to set currency is sent, are you sure in that case it will work too? this process is not related to set currency page

nazaninreihani: @brucebinary the process is like this =>
please choose account currency in 'create new account' in user accounts section and go to upgrade (upgrade via maltainvest account opening), then upgrade, this should work correct, I'll test it too tommorow

brucebinary:
@nazaninreihani Here's the PR to fix the set currency issue: https://github.com/binary-com/binary-next-gen/pull/756 ...though, I'm not sure if it refers to the same issue. Cause I was able to set the currency via the set currency page just fine.

nazaninreihani:
@brucebinary after setting the currency it's added to localStorage but due to removing the refresh from this response seems that app still thinks that user has not set currency and shows him the set currency for this account button and does not let him to add a new account

PR #756, opened 2017-12-11
bug fix: currency not set when adding real account

bruce-binary:
This only occurs in instances where you did not change the currency input when adding a real account. This commit selects the first available currency, should it becomes available.

### DIFF
--- a/src/_data/LiveData.js
+++ b/src/_data/LiveData.js
@@ -253,12 +253,15 @@ export const trackSymbols = symbols => {
     api.subscribeToTicks(symbols);
 };
 
-export const setAccountCurrency = (currency, store) => {
+export const setAccountCurrency = (currency, store, onCurrencyAdded) => {
   const state = store.getState();
   api.setAccountCurrency(currency).then(r => {
       if (r.set_account_currency === 1) {
         const loginid = state.account.toJS().loginid;
         addCurrencyToAccount(currency, loginid);
+        if (onCurrencyAdded) {
+          onCurrencyAdded();
+        }
       }
     }
   );
--- a/src/_data/LiveData.js
+++ b/src/_data/LiveData.js
@@ -172,36 +172,38 @@ const initAuthorized = async (authData, store) => {
                 allowedMarkets = landingCompany.financial_company.legal_allowed_markets;
             }
             canUpgrade = !hasAccountOfType('real', accounts);
-        } else if (landingCompany.financial_company.shortcode === 'maltainvest') {
-            typeOfNextAccount = 'financial';
-            canUpgrade = !hasAccountOfType('financial', accounts);
-            currencyOptions = landingCompany.financial_company.legal_allowed_currencies;
-            allowedMarkets = landingCompany.financial_company.legal_allowed_markets;
-        } else if (landingCompany.financial_company.shortcode === 'costarica') {
-            allowedMarkets = landingCompany.financial_company.legal_allowed_markets;
-            const legalAllowedCurrencies = landingCompany.financial_company.legal_allowed_currencies;
-            const existingCurrencies = getExistingCurrencies(accounts);
-            if (existingCurrencies.length) {
-                const dividedExistingCurrencies = groupCurrencies(existingCurrencies, currencyConfig);
-                if (dividedExistingCurrencies.fiatCurrencies.length) {
-                    const dividedAllowedCurrencies = groupCurrencies(legalAllowedCurrencies, currencyConfig);
-                    const legalAllowedCryptoCurrencies = dividedAllowedCurrencies.cryptoCurrencies;
-                    const existingCryptoCurrencies = dividedExistingCurrencies.cryptoCurrencies;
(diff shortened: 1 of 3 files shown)


src/user-accounts/CreateNewAccount.js
```diff
@@ -31,14 +31,21 @@ export default class CreateNewAccount extends PureComponent {
 
   onRedirectToAccountOpening = () => {
     this.setState({ currency_error: false });
-    if (this.props.account.currency && this.props.account.currency !== '') {
+    if (this.state.selected_currency && this.state.selected_currency !== '') {
```
brucebinary: @nazaninreihani this is what you intended right? Cause `this.props.account.currency` will return the currency of the active account.

brucebinary:
@nazaninreihani does this mean that we are going to remove the "upgrade" button? We can't select the currency from there, so if the customer upgrades from there they might get confused on the currency error message during trading. 

The error message that appears during trading should direct them to select a currency, or we upon upgrade we redirect them to the set currency page.

src/user-accounts/CreateNewAccount.js
nazaninreihani: @brucebinary I fixed it here https://github.com/binary-com/binary-next-gen/pull/754/files but this PR seems better to me, 
but the issue I was talking about is not this, the issue is when user chooses a currency in set-currency page or for upgrading, it sends the request to set that currency and in response we should add the currency to account in localStorage, but by moving the refresh code from inside the function which is going to run after setting currency, it's not going to refresh after the response so the currency is added to localStorage but still the app thinks the user has no currency.

brucebinary: @nazaninreihani I'll take a look at that. I just tried setting the currency this morning, and I didn't come across any issues.

nazaninreihani:
@brucebinary In binary.com we still have both of these ways for upgrading, I'm not sure about removing upgrade button, also I'm adding a notification section to app so user will see they need to choose currency too.
but let's ask Arnab & Amin about it

brucebinary:
@arnabk a note: https://github.com/binary-com/binary-next-gen/pull/756/commits/cced6bd6b6396bda935d8a79dfb996cf9e449117 and https://github.com/binary-com/binary-next-gen/pull/754/commits/0667d28574347231b55dea1dc8395b4fb7220a80 fixes the same issue. To solve the problem where the available currencies arrived at a delayed time, the former sets the currency state value the moment the currencies are successfully retrieved, and the latter sets the currency by retrieving the first value in the retrieved currencies (assuming by the time the user submits the form, the currencies would have already been loaded).

arnabk:
@nazaninreihani @brucebinary I think you two know more on what needs to be done than me. @ashkanx knows more about the functional side of it. May be he can comment too.

What I have noticed that binary-static & webtrader both shows a notification bar to set the default currency. Ideally that is how it should be for next-gen too.

src/_data/LiveData.js
```diff
@@ -172,36 +172,38 @@ const initAuthorized = async (authData, store) => {
...
-            const existingCurrencies = getExistingCurrencies(accounts);
-            if (existingCurrencies.length) {
-                const dividedExistingCurrencies = groupCurrencies(existingCurrencies, currencyConfig);
-                if (dividedExistingCurrencies.fiatCurrencies.length) {
-                    const dividedAllowedCurrencies = groupCurrencies(legalAllowedCurrencies, currencyConfig);
-                    const legalAllowedCryptoCurrencies = dividedAllowedCurrencies.cryptoCurrencies;
-                    const existingCryptoCurrencies = dividedExistingCurrencies.cryptoCurrencies;
-                    currencyOptions = legalAllowedCryptoCurrencies.filter(x => existingCryptoCurrencies.indexOf(x) === -1);
-                    if (currencyOptions.length) {
+        } else if (landingCompany.financial_company) {
```
brucebinary: @nazaninreihani in MLT user cannot set currency because unable to find `shortcode` of `undefined`, so I added this small check here, but I'm not sure if it will cover all use cases? It works fine here.

brucebinary: hm. Should have made a new PR, but it is the same issue name so...

nazaninreihani: @brucebinary you mean when you tried to upgrade from MLT to MF via create new account the select box of currency was empty?

nazaninreihani:
@arnabk @brucebinary I was working on a notification section but for now I paused it, maybe I'd continue it in few days

src/_data/LiveData.js
brucebinary: My bad for the confusion. The error happens when user that can only have MLT accounts (Belgium).

nazaninreihani: @brucebinary I see it now, I think this PR will fix it, thanks

brucebinary:
@arnabk On the 2 PR's that fix the same issue: I've discussed with @nazaninreihani and we choose this PR. Should be green light to merge.

PR #760, opened 2017-12-12
fix error of default currency (happens only in slower connections, it…

nazaninreihani:
… throws error when default currency in state is an empty object either empty string, but with setting it to default USD it works)


brucebinary:
This doesn't just happen in slower connections. QA said the errors caused the reality check to be unable to close.

PR #765, opened 2017-12-14
when adding CR account from CR account, bypass the upgrade page

bruce-binary:


src/user-accounts/CreateNewAccount.js
```diff
@@ -19,36 +23,74 @@ export default class CreateNewAccount extends PureComponent {
...
+      const response = await api.createRealAccount(createAccountParams);
+      addNewAccount(response.new_account_real);
+      location.reload();
+    } catch (e) {
+      this.setState({ currency_error: true });
+      this.setState({ disable_create_account: false });
+    }
+  }
+
   onRedirectToAccountOpening = () => {
```
nazaninreihani: @brucebinary Looks Good, thanks, 
I think with this behavior better to rename this function to submitCreateAccount or something that makes sense more, what do you think?

nazaninreihani (approved):

src/user-accounts/CreateNewAccount.js
brucebinary: @nazaninreihani I think `submitCreateAccount` is too general, since the function is specifically used to add a CR account only when there is an existing CR account. How about `submitCreateCRAccount`?

sry for the late reply.

brucebinary:
@arnabk hi we have agreed on the refactoring. Green light to merge

PR #775, opened 2017-12-18
bug fix: Drop down "Address state" is empty on upgrade form.

bruce-binary:
Issue is more evident when there are more states (i.e. Greece)

Also there's a mix of tabs and spaces. Would be nice if it was consistent. Can we agree on either one?


arnabk:
@brucebinary No test classes for this change?

brucebinary:
You prolly need selenium to test this. This is an intermittent that bug only surfaces with large number of address states or slow internet speed, where there is a slight delay for the server to send over the data. I don't think we can write unit tests for this. But @nazaninreihani would know better.

arnabk:
@nazaninreihani Please review and let me know if we could merge.
@brucebinary There are merge conflicts.

brucebinary:
@arnabk I just merged with the latest dev branch. Turns out this issue is already fixed by @nazaninreihani. It's still worth merging tho: in the maltainvest card, currently the 1st address state is the default selected. We want the default value selected to be empty.

nazaninreihani:
@arnabk @brucebinary I wanted to add componentWillReceiveProps and I thought if Bruce added it in his PR so I checked and I found out he added that, so I copied his code and changed it in the way to fix another issue, I thought that his PR will be merged and after that I will resolve conflicts and my changes will be merged, my bad I shouldn't have submitted the PR before this PR, I'm sorry @brucebinary

nazaninreihani (approved):

PR #778, opened 2017-12-19
When there are validation errors, user should be scrolled to field wh…

bruce-binary:
…ere error is present.

There's no scrolling animation here. That's not easy to do unless you use jQuery or some other animation library, which would add to the size of the app. For now, just this will do.  At least the user will know where the validation error is than the page just not progressing.


arnabk:
@brucebinary Is https://reactjs.org/docs/react-dom.html#finddomnode the recommended approach than using document directlty?

brucebinary:
In this context, no. We want to access the element by ID, not search the react DOM for a specified component.

arnabk:
document.getElementById is not a free call. It is a search too. I have not seen any article recommend mix of document in React components. Could you please point me to the right article where I could and find this as a recommended practice. I still think that we should be use this.refs.<something> instead of using document.getElement

brucebinary:
@arnabk so I did my research a bit. Yes, you are right. @aminroosta pointed out that retrieving the  directly from the HTML DOM is an anti-pattern because the whole point of using React is to delegate all DOM operations to React.

What `findDomNode()` does is that given a react component (referenced by `ref='uniqueName'`) it returns the corresponding node in the HTML DOM tree. To implement this it would require us to add a string ref `ref='id-name'` (in addition to already having `id='id-name'` to every single `InputGroup` component.

As tedious as it may sounds, it is still not a good solution looking long term. As noted in the note section of the [findDomNode](https://reactjs.org/docs/react-dom.html#finddomnode) documentation, the standard practice is not to use `findDomNode` at all, because the React team [plans to remove string refs in the future](https://reactjs.org/docs/refs-and-the-dom.html#legacy-api-string-refs). What is the recommended practice by the React folks? [Exposing DOM Refs to Parent Components](https://reactjs.org/docs/refs-and-the-dom.html#exposing-dom-refs-to-parent-components).

I check up on an example from mozilla voice on how to do this, and for something seemingly so simple [it looks very tedious](https://github.com/mozilla/voice-web/pull/568/files#diff-e4a4c90de0fea7aba80f420b62a00331R189). Imagine, for each `InputGroup`, you have to add a callback that stores its component reference to a corresponding variable in the parent component (alternatively use one hash table that stores all these references). 

Anyhow. I'll update this PR to use refs instead.

PS: document.getElementById use O(1), since it is just doing a hash table lookup: https://stackoverflow.com/a/10444121/1471258

src/settings/SettingsSelfExclusion.js
```diff
@@ -73,7 +73,7 @@ export default class SettingsSelfExclusion extends PureComponent {
     this.setState({ errors: newErrors });
     const keys = Object.keys(newErrors);
     if (keys.length > 0) {
-      document.getElementById(keys[0]).scrollIntoView();
+      document.getElementById(keys[0]).scrollIntoView({ block: 'center', behavior: 'smooth' });
```
brucebinary: lol turns out it is supported in chrome also

src/settings/SettingsSelfExclusion.js
```diff
@@ -88,6 +91,7 @@ export default class SettingsSelfExclusion extends PureComponent {
         success: true,
         serverError: false,
       });
+      document.getElementsByClassName('update-notice')[0].scrollIntoView();
```
brucebinary: @arnabk @aminroosta guys. I have here a preview of using refs instead of getElementById: https://github.com/brucebinary/binary-next-gen/compare/scroll-to-error-fix...brucebinary:scroll-to-error-finddomnode My point is to demonstrate that using refs in this context is tedious and difficult to understand, for something that is seemingly trivial. What is happening is that the we are still accessing the DOM node, but instead of doing it directly (react components don't have a `scrollIntoView` method, unless you install a library for it) via `document` we access it through the corresponding react component via `findDOMNode` (use of `findDOMNode` is actually ill-advised and eslint will throw errors for it). 

However, if everyone agrees to do it this way I will update this PR accordingly.

However I'm a little stuck with the above line of code. What it does it scroll to the update notice when the settings have successfully updated. Should we decide to use refs, the ref to this `UpdateNotice` component will need to be passed to this component, in which we again use it to find the corresponding DOM node via `findDOMNode` and call `scrollIntoView`.

Bear in mind all we want to do is to scroll the view to a specified input field.

src/settings/SettingsSelfExclusion.js
```diff
@@ -88,6 +91,7 @@ export default class SettingsSelfExclusion extends PureComponent {
         success: true,
         serverError: false,
       });
+      document.getElementsByClassName('update-notice')[0].scrollIntoView({ block: 'center', behavior: 'smooth' });
```
brucebinary: @arnabk @aminroosta guys. I have here a preview of using refs instead of getElementById: https://github.com/brucebinary/binary-next-gen/compare/scroll-to-error-fix...brucebinary:scroll-to-error-finddomnode My point is to demonstrate that using refs in this context is tedious and difficult to understand, for something that is seemingly trivial. What is happening is that the we are still accessing the DOM node, but instead of doing it directly (react components don't have a `scrollIntoView` method, unless you install a library for it) via `document` we access it through the corresponding react component via `findDOMNode` (use of `findDOMNode` is actually ill-advised and eslint will throw errors for it). 

However, if everyone agrees to do it this way I will update this PR accordingly.

However I'm a little stuck with the above line of code. What it does it scroll to the update notice when the settings have successfully updated. Should we decide to use refs, the ref to this `UpdateNotice` component will need to be passed to this component, in which we again use it to find the corresponding DOM node via `findDOMNode` and call `scrollIntoView`.

Bear in mind all we want to do is to scroll the view to a specified input field.

arnabk: @brucebinary Couple of comments

```
const addToFormDOM = inp => { if (inp) this.DOMNodes[inp.props.id] = ReactDOM.findDOMNode(inp); };
```
It might be better not to have this function inside render method.
Try ` if (inp) this.DOMNodes[inp.props.id] = inp;`. Not sure if it will work, but worth a try.

https://github.com/stipsan/scroll-into-view-if-needed
Is it at all useful to use this library? Not sure if it would help us in future to scroll differently with different types of animation.

Just a thought. Other than that, I think these changes are as per React patterns, although I do hear your point that just to do the pattern, we are writing double the code.

aminroosta:
@arnabk I think we can use `document.getElementById` in this case,
 react is just not mature enough to handle scrolling natively, following the `ref` convention doesn't add any tangible improvement either.

I'd say we just use `getElementById` since I love oneliners so much :-)

src/settings/SettingsSelfExclusion.js
brucebinary: @arnabk the code `if (inp) this.DOMNodes[inp.props.id] = inp;` will not work because `inp` is a react component. What we want is the HTML DOM node.

The scroll library linked also requires the HTML DOM to be passed as argument. It doesn't take a react component. The library I was referring to is this: https://github.com/flyingant/react-scroll-to-component 

We will still need that `addToFormDOM` helper method, but at least we won't use `React.findDOMNode`.

arnabk:
@brucebinary @aminroosta I do not want to be pain in a** here. I believe it is upto the developer to make the best call. My personal experience working in large scale React based project is that, if you use $ or VanillaJS in one place, then new comers replicate it in so many places that it is hard to stop later. You cannot expect all members of the team to be at the highest level of knowledge. But I do not understand @brucebinary 's point here. Please fix the conflict and I will go ahead and merge it.

